PR #1, opened 2020-04-07
Allow only a certain set of PR actions to trigger the run

orta:
Updates some logs and adds an allsowlist of events which can trigger the run


src/_tests/fixtures/43314/result.json
```diff
@@ -1,6 +1,6 @@
 {
   "pr_number": 43314,
-  "targetColumn": "Waiting for Code Reviews",
```
orta, line 3: Fixtures don't contain the date when they were recorded, so this one hit the "greater than 5 days" check and moved into maintainer review

PR #8, opened 2020-04-08
Don't needlessly edit posts to the same content

orta:
Should make it easier for us to see the differences in issues


orta:
I'd be shocked if this breaks something

Issue #26, opened 2020-04-29
Delete old travis update messages


orta:
These are probably force pushes - https://github.com/DefinitelyTyped/DefinitelyTyped/pull/44299

orta:
Fixed in #31

PR #29, opened 2020-04-29
Flatten the PR mutations code

orta:
Refactor: Moves each part of the mutations to be their own function. I added two extra vars for 'add labels' and 'change project columns' _in anticipation_ of #26  - but this just changes the shape of that file.


orta:
The mutations are re-ordered due to the concats, but they all seem to be there

src/comments.ts
```diff
@@ -26,7 +26,7 @@ export const MergeConflicted = (slug: string, user: string) => ({
 });
 
 export const ChangesRequest = (headOid: string, user: string) => ({
-    tag: `complaint-${headOid}`,
+    tag: `reviewer-complaint-${headOid}`,
     status: `@${user} One or more reviewers has requested changes. Please address their comments. I'll be back once they sign off or you've pushed new commits or comments. If you disagree with the reviewer's comments, you can "dismiss" the review using GitHub's review UI. Thank you!`
 });
```
orta, line 31: This is never used for looking up stuff, but at least now they are all scoped according to their domain

PR #31, opened 2020-04-29
Remove old travis comments

orta:
You only care about https://github.com/DefinitelyTyped/dt-mergebot/commit/7d11db22bc245cf4216c81840426bb306b9f6969

Commenting inline


src/execute-pr-actions.ts
```diff
@@ -153,6 +157,26 @@ async function getMutationsForComments(actions: Actions, pr: PR_repository_pullR
...
+    if (!parsed) continue
+
+    if (parsed.tag.includes("travis") && parsed.tag !== travisMessageToKeep?.tag) {
+      mutations.push( createMutation(deleteComment, { input: { id: comment.id } }) )
+    }
+  }
+
+  return mutations;
+}
+
```
orta, line 179: Basically find all bot comments about travis which aren't the same as a potential message we might be posting

andrewbranch:
What’s the motivation for this exactly? Is this to clean up from something that went wrong, or is it useful long-term? Can’t quite identify the scenario.

orta:
I don't think we should be leaving a trail of useless comments about old travis results behind in a PR mainly

orta:
I'm gonna merge, if we think it's annoying then a less `--force` approach to is to minimize comments instead. With danger it got annoying real fast when making a new comment each time.

PR #35, opened 2020-04-30
Add support for pinging reviewers and not on mega PRs

### LINKED ISSUE #28: What should we do with mega-PRs?
E.g. https://github.com/DefinitelyTyped/DefinitelyTyped/pull/44272

e.g. I don't think we should be pinging almost all contributors - is there a 'this PR affects too many packages' limit?

### LINKED ISSUE #34: The bot doesn't ping dt module owners
Doh! Examples:

- https://github.com/DefinitelyTyped/DefinitelyTyped/pull/44365
- https://github.com/DefinitelyTyped/DefinitelyTyped/pull/44364

orta:
Adds the ping to actually get people to review! Fixes #34 

Takes into account mega PRs which fixes #28


src/compute-pr-actions.ts
```diff
@@ -70,6 +70,16 @@ export function process(info: PrInfo): Actions {
         status: createWelcomeComment(info)
     });
 
+    // Ping reviewers
+    if (!info.anyPackageIsNew && info.owners.length) {
```
andrewbranch, line 74: Why `!info.anyPackageIsNew`? You could have a PR that adds a new package but also touches packages with owners that should be pinged, right?

orta: Yeah, that's reasonable - I think the length should be enough in that case then

PR #52, opened 2020-05-01
Improve status fail logging

orta:
I'm not certain why we're not getting PR details - so this prints the query used to find the PR for a commit


orta:
There's a chance that it's from someone force-pushing and overwriting a commit, which isn't really something we can handle without skipping the run

orta:
Was the error
Screen Shot 2020-05-01 at 11 53 24 AM

PR #61, opened 2020-05-01
Improve the travis status detection

orta:
We had two ways to detect travis results: 

- checksuites
- status updates

The check suites came first, and it'd very rarely hit the 2nd which was more accurate. So, I removed the lookup in the check suite and always use the status update search.


src/_tests/fixtures/43695-post-review/result.json
```diff
@@ -21,11 +21,15 @@
...
     },
     {
       "tag": "reviewer-complaint-90c94f9",
       "status": "@alexandercerutti One or more reviewers has requested changes. Please address their comments. I'll be back once they sign off or you've pushed new commits or comments. If you disagree with the reviewer's comments, you can \"dismiss\" the review using GitHub's review UI. Thank you!"
+    },
+    {
+      "tag": "stale-ping-fbf110-a5285cd",
+      "status": "@RyanCavanaugh Thank you for reviewing this PR! The author has pushed new commits since your last review. Could you take another look and submit a fresh review?"
     }
```
orta, line 33: This PR used to be classed as red - now it's green

src/pr-info.ts
```diff
@@ -479,49 +479,24 @@ function getTravisResult(headCommit: PR_repository_pullRequest_commits_nodes_com
...
-                    travisUrl = totalStatusChecks.targetUrl;
-                    break;
-                
-                    case StatusState.EXPECTED:
-                case StatusState.PENDING:
-                default:
-                    travisResult = TravisResult.Pending;
-                    break;
-            }
-        }
```
orta, line 524: The TDLR is I moved this up and removed the top `case StatusState.PENDING` which was wrong

PR #63, opened 2020-05-01
Adds unpkg urls to new deps

### LINKED ISSUE #58: Add unpkg link to welcome comment for new packages
The npm link is useful for gauging popularity, but I also use unpkg on every new definition review to check module format.

orta:
The unpkg link takes you to the root folder and it's up to you top get to the main JS export, we could move to have a few API calls in here to eventually make that easier to give a single click 'load the main js files'.

fixes #58


orta:
fixes #58

PR #69, opened 2020-05-04
Make travis result/failing review results mutually exclusive

weswigham:
Which should hopefully fix categorization of PRs like https://github.com/DefinitelyTyped/DefinitelyTyped/pull/44373 which have requested changes, but were erroneously marked as "Needs Maintainer Review", rather than "Needs Author Action".


andrewbranch:
Apparently this bug was already encoded into a test baseline. You can run `npm run update-all-fixtures` to overwrite them.

PR #75, opened 2020-05-08
Be laxer on who can comment and keep the PR alive

orta:
Originally this was just the PR author, but now it can be anyone.

PR is noisy because I changed the name of an info variable, scroll right to the bottom to see code changes


src/pr-info.ts
```diff
@@ -238,7 +238,7 @@ export async function deriveStateForPR(
         stalenessInDays: daysSince(headCommit.pushedDate, now),
         lastCommitDate: lastPushDate,
         reopenedDate: getReopenedDate(prInfo.timelineItems),
-        lastAuthorCommentDate: getLastAuthorActivityDate(prInfo.author.login, prInfo.timelineItems, prInfo.reviews) || lastPushDate,
+        lastCommentDate: getLastCommentishActivityDate(prInfo.timelineItems, prInfo.reviews) || lastPushDate,
```
andrewbranch, line 241: Related, but not a blocker for this PR: I actually had a mental review to stop using `timelineItems` for getting comment dates, because IIRC we already fetch the last 100 comments separately on the PR. We may be able to cut down significantly on the query cost and response size that way.

orta: 👍 true

PR #85, opened 2020-05-15
When a PR goes from auto mergable to not - remove the comment offering to merge the PR

orta:
We just had a big PR which made a lot of PRs get merge conflicts, and a few have now gone from "ok, I can merge this for you" back to "this needs a rebase, then reviews" - we need to kill the message saying it can merge.


src/compute-pr-actions.ts
```diff
@@ -40,6 +40,7 @@ export interface Actions {
     shouldUpdateLabels: boolean;
     shouldUpdateProjectColumn: boolean;
     shouldRemoveFromActiveColumns: boolean;
+    isClassedAsReadyForMerge: boolean
```
sandersn, line 43: I assume prettier is going to add a semicolon here to match the other lines (or perhaps...remove the others?)

src/compute-pr-actions.ts
```diff
@@ -121,6 +124,7 @@ export function process(info: PrInfo | BotEnsureRemovedFromProject | BotNoPackag
     context.labels["New Definition"] = info.anyPackageIsNew;
     context.labels["Author is Owner"] = info.authorIsOwner;
     context.labels["Merge:Auto"] = canBeMergedNow(info);
+    context.isClassedAsReadyForMerge = canBeMergedNow(info);
```
sandersn, line 127: so the problem was that `canBeMergedNow` wasn't persisted before?

src/compute-pr-actions.ts
andrewbranch:
```suggestion
    isMergeable: boolean;
```

?

“classed as” feels very clunky

orta:
@andrewbranch how dare you defy up my old school cocoa naming structure

That said, maybe "isReadyForAutoMerge" is probably the best (is mergeable could be that the PR is in a mergeable state)

src/compute-pr-actions.ts
orta: Nah, I've not set up prettier for this repo - some of the older code would probably look pretty bad so I didn't add it after I took over

src/compute-pr-actions.ts
orta: effectively it wasn't being passed down to the 'commenting' stage and couldn't be used to show/remove that message

PR #91, opened 2020-05-21
Use the access rights function to determine which project column to place the PR

### LINKED ISSUE #88: A PR without tests which has reviews from a module owner should be in 'maintainer review'

orta:
Fixes #88


src/compute-pr-actions.ts
```diff
@@ -234,10 +235,12 @@ function canBeMergedNow(info: PrInfo): boolean {
     return hasFinalApproval(info).approved;
 }
 
+type PotentialReviewers = "DT maintainers" | "type definition owners or DT maintainers" | "type definition owners, DT maintainers or others"
```
orta, line 238: Ensures that we can rely on the strings in the function below

PR #95, opened 2020-06-01
Allow check suite webhooks to update the bot

orta:
We used to use travis but #94 changed that, and there are 2 systems for getting red/green information on a PR:

- statuses
- checks

Statuses are the old way (and how travis does it), they are ad-hoc red/yellow/green messages that get attached to a commit. Checks on the other hand are structured data where you connect the PR and the commit. The Checks API is that yo make a suite per CI run and you can have many checks inside that suite, then you say it has ended. Azure/GH Actions use this.

Basically we were listening for status updates to look over the finished PR because travis used that - but now we need to look for the completion of a check suite.


PR-Trigger/index.js
```diff
@@ -6,7 +6,6 @@ const {executePrActions} = require("../bin/execute-pr-actions")
 const {mergeCodeOwnersOnGreen} = require("../bin/side-effects/merge-codeowner-prs")
 const verify = require("@octokit/webhooks/verify");
 const sign = require("@octokit/webhooks/sign");
-const {runQueryToGetPRMetadataForStatus} = require("../bin/queries/status-to-PR-query")
```
orta, line 9: because a status was disconnected from a PR, you would use the GH API to search for a PR containing that commit - hacky, but worked, we don't need this now

PR #102, opened 2020-06-22
Adds `No Other Owners` label when appropriate (& not for new packages)

elibarzilay:
Note that it moves `const otherOwners` higher, so it can be used by the
label code.


orta:
Cool, makes sense 👍

PR #108, opened 2020-06-29
Update packages

elibarzilay:
I tried this in an attempt to debug an apollo failure, it didn't help
but I figured that it might be good to keep it.  Gets rid of all current
audit problems too.

All done via `npm` so should be fine, I think.


orta:
Sure, if you've tested it :+1:

elibarzilay:
Yeah, worked perfectly fine when I ran it locally.  I'll merge and we'll see if the server has a different opinion...

PR #111, opened 2020-07-01
`.gitignore`: fix and add `coverage`

elibarzilay:
(`coverage` is for the generated jest coverage reports.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
-bin/
-node_modules/
+/bin/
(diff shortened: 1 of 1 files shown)


orta:
Sure thing, but I'll be resistant to adding code coverage metrics to this repo in a way that breaks builds 👍

elibarzilay:
@orta, agreed -- just making it easier to run it locally

PR #112, opened 2020-07-01
`_tests/fixturedActions.test.ts`: lint, reorganize, tweak

elibarzilay:


elibarzilay:
Yeah, because I moved a chunk of code -- but there are no actual functionality changes in this.

PR #119, opened 2020-07-02
Simplify workflow: drop the 4 day thing

elibarzilay:
There's no need for it now that good PRs are subject to being abandoned
and then they'll move to the maintainer queue anyway.

@RyanCavanaugh made a point about not waiting too much to avoid
conflicts -- if that's a desired point, then I'll abstract the
abandonment timeline into one for bad prs and a shorter one for good
ones.


orta:
I'm cool with the simplification, but from a quick look this moves the time when we look at a PR with no peer feedback from 4 days to 7, we should ideally keep that short - so if we switch the idleness down to 4 I'm good

elibarzilay:
Closing this, since this turns into a different pr for reviving YSYL.

PR #123, opened 2020-07-07
Show `remove` results in the output

elibarzilay:
Makes it easier to see which PRs are ignored because they're draft PRs.


orta:
Cool idea

PR #124, opened 2020-07-07
Random improvements in `pr-info.ts`

elibarzilay:
`categorizeFile` uses one regexp to analize the path, and makes a
decision based on the suffix which should be easier to extend if needed,
including grabbing the subpath (e.g., to require `test` in the name of
test files).

Use a `Set` for `getPackagesTouched` just because I was waiting...

### DIFF
--- a/src/pr-info.ts
+++ b/src/pr-info.ts
@@ -319,32 +319,19 @@ type FileLocation = ({
 }) & { filePath: string };
 
 function categorizeFile(filePath: string): FileLocation {
-    const typeDefinitionFile = /^types\/([^\/]+)\/(.*)\.d\.ts$/i;
-    // https://regex101.com/r/QfAfRn/1
-    const typeTestFile = /^types\/([^\/]+)\/(.*)\.tsx?$/i;
-    const typeOtherFile = /^types\/([^\/]+)\/(.*)$/i;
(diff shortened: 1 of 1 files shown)


src/pr-info.ts
```diff
@@ -319,32 +319,18 @@ type FileLocation = ({
...
-    let match;
-    if (match = typeDefinitionFile.exec(filePath)) {
-        return { filePath, kind: "definition", package: match[1] };
-    } else if (match = typeTestFile.exec(filePath)) {
-        return { filePath, kind: "test", package: match[1] };
-    } else if (match = typeOtherFile.exec(filePath)) {
-        return { filePath, kind: "package-meta", package: match[1] };
-    } else {
-        return { filePath, kind: "infrastructure" };
+    const match = /^types\/(.*?)\/.*?[^\/](?:\.(d\.ts|tsx?))?$/.exec(filePath);
```
orta, line 322: Can you include a link like: `https://regex101.com/r/QfAfRn/1` which shows the regex with some passing and failing tests please

elibarzilay: Done, though for some reason the highlights on this thing are broken when I switch to JS mode.  But the match info on the RHS is fine, so I left it in.

PR #126, opened 2020-07-07
Add the maintainer blessing feature

elibarzilay:
This is implemented by moving the PR to a different column.  There's one
big caveat though: *any* column moving (even back into the maintainer
view) is considered as blessing the PR (and therefore there is no way to
cancel the blessing).  This is because there is no info from the GH api
about the from/to column names -- there is a `previousProjectColumnName`
field which is in beta now, and when it's available, then this should be
extended so that a blessing is a move out of the main queue column.

Note that this commit adds the blessing detection + the info field, but
doesn't do anything with it yet.

* Add a `maintainerBlessed` field.

* Extend the `TimelineItems` query to grab `MovedColumnsInProjectEvent`
  events.

* Extended the timeline query to 200 items (the limit is 250).

* Extend `authorNotBot` to work on nodes with an `actor.login`.


src/pr-info.ts
```diff
@@ -241,6 +254,7 @@ export async function deriveStateForPR(
         stalenessInDays: Math.min(...[lastCommitDate, lastCommentDate, reopenedDate, reviewAnalysis.lastReviewDate]
                                      .map(date => daysSince(date || lastCommitDate, now))),
         lastCommitDate, reopenedDate, lastCommentDate,
+        maintainerBlessed: lastBlessing ? lastBlessing.getTime() > lastCommitDate.getTime() : false,
```
weswigham, line 257: Commit times can be manufactured artificially by the uploader and are not reliable - so using them for access control related things seems sketchy. Our mindset working here very much needs to be that the committer on a PR is a potentially hostile actor seeking to damage DT in some way, and that's what we seek to guard against (by only allowing changes we know to be safe/valid).

orta: I agree in general, though in this case, manipulating your commit dates will only revert behavior back to letting the bot decide which column to place and to my knowledge that should not affect access control. I guess it depends on where Eli takes it from here.

elibarzilay: I verified that this is not a problem: the code uses `headCommit.pushDate`, so I force-pushed an update to this PR with fake author and commit dates.  Using the GH api, all three dates are visible:

```
"pushedDate": "2020-07-07T17:05:59Z",
"authoredDate": "2021-07-07T17:00:00Z",
"committedDate": "2019-07-07T17:00:00Z"
```

and we're using the first one.

weswigham: ah, cool, it's just the local variable name that's misleading then, nice~

PR #129, opened 2020-07-07
Refactor `hasFinalApproval` in a functional way

elibarzilay:
Still works the same, but will make the following changes easier to follow.


elibarzilay:
@orta: there's just one more level of indirection here -- the bottom part decides which level is needed (specified as `others`, `owners`, or `maintainers`), and then the top `approvalFor` constructs the result based on that level and the approval bits.

The main thing is that this makes it easy to add the `|| info.maintainerBlessing` bit later on to the top, rather than threading it throughout the current conditional branches.

orta:
Sure, let's have it in

PR #130, opened 2020-07-08
Detect markdown files too

elibarzilay:
* `FileLocation` -> `FileInfo`, and unified by making `package` possibly
  `undefined`.  Also: `filePath` -> `path` (would be easier to make it
  extend the info from the api if the additions and deletions are useful
  in the future).

* Simplify `getDangerLevel` unsing `.some` instead of a single for loop.

* Added as a new kind, which will be ignored by `getDangerLevel` so
  won't count as a `ScopedAndConfiguration`.


src/pr-info.ts
```diff
@@ -335,32 +335,28 @@ function getLastMaintainerBlessingDate(timelineItems: PR_repository_pullRequest_
...
+function categorizeFile(path: string): FileInfo {
+    // https://regex101.com/r/eFvtrz/1
+    const match = /^types\/(.*?)\/.*?[^\/](?:\.(d\.ts|tsx?|md))?$/.exec(path);
+    if (!match) return { path, kind: "infrastructure", package: undefined };
     const [pkg, suffix] = match.slice(1); // `suffix` can be null
-    switch ((suffix || "").toLowerCase()) {
-        case "d.ts": return { filePath, kind: "definition", package: pkg };
-        case "ts": case "tsx": return { filePath, kind: "test", package: pkg };
-        default: return { filePath, kind: "package-meta", package: pkg };
+    switch ((suffix || "")) {
```
orta, line 351: Did you want to lose the `.toLowerCase`?

src/pr-info.ts
```diff
@@ -335,32 +335,28 @@ function getLastMaintainerBlessingDate(timelineItems: PR_repository_pullRequest_
...
+type FileKind = "test" | "definition" | "markdown" | "package-meta" | "infrastructure";
+
+type FileInfo = {
+    path: string,
+    kind: FileKind,
+    package: string | undefined
+};
+
+function categorizeFile(path: string): FileInfo {
+    // https://regex101.com/r/eFvtrz/1
```
orta, line 347: Thannkkkkkssss

src/pr-info.ts
```diff
@@ -446,52 +442,24 @@ function analyzeReviews(prInfo: PR_repository_pullRequest, isOwner: (name: strin
     });
 }
 
-function getDangerLevel(categorizedFiles: readonly FileLocation[]) {
+function getDangerLevel(categorizedFiles: readonly FileInfo[]) {
     if (categorizedFiles.some(f => f.kind === "infrastructure")) {
         return "Infrastructure";
+    }
+    const packagesTouched = getPackagesTouched(categorizedFiles);
+    if (packagesTouched.length === 0) { // ????
```
orta, line 450: If you edit the root `README.md` or localization PRs etc

src/pr-info.ts
elibarzilay: Yes: since the regexp is not case-insensitive anyway, so it didn't do much.  I considered adding an `i` flag, but looking at the repo there was just one file with an uppercase `.MD` which I renamed earlier, so no reason to allow weird casing now...

src/pr-info.ts
elibarzilay: I know -- that `???` comment was there before, I just moved it up.  (It didn't seem questionable to me at all, so I figured that someone had a reason to put it there...)

PR #131, opened 2020-07-08
Consider only the last review from each author

elibarzilay:
The code was already ignoring stale reviews for a different
`headRefOid`, but it might be that a second review approves a request
without changes, so the first review for the same author should be
ignored even if it's the same headref.

(Note that the test case still doesn't want to merge it, because the PR
also edited multiple packages.)


orta:
Cool, yeah, makes sense 👍

PR #132, opened 2020-07-08
Scrub diagnostics in markdown in json test files

### LINKED ISSUE #127: Strip out the summary JSON in the markdown of the fixtures?
This would mean PRs have smaller diffs in general, and knowing that the JSON is in there is never actually something we're testing  in reality.

elibarzilay:
It repeats information in `derived.json`, just in a harder to inspect
way due to long line of text.  The regexp that is used is careful to
delete only these things (there are more uses of `<details>` for the
benchmark results and for many reviewers).

Done both in the testing code and in `create-fixture.ts`.  I also
removed it manually from existing input files to keep things compact.

(And also some more hand linting.)

Closes #127.


orta:
Looks good, nice one

PR #136, opened 2020-07-12
Add the ability to fetch meta files and test them

elibarzilay:
Includes simple tests for valid `OTHER_FILES.txt` and `tslint.json`,
marking them as `package-meta-ok` which is ignored for the decision to
put a PR in the maintainer queue.


elibarzilay:

PR #139, opened 2020-07-12
Enable maintainer blessings

elibarzilay:
* Add a `Needs Maintainer Action` column for PRs that needs an action
  and cannot be dismissed by blessing, make the relevant cases use it
  instead of `Needs Maintainer Review`.

* Switch language like "DT maintainer needs to merge changes" to "... to
  approve changes" for cases where the maintainer can bless the PR and
  the merging will then be done by the author.

* Do not require maintainer approval if a pr would go to `Needs
  Maintainer Review` and it's blessed.

* `info.dangerLevel === "Infrastructure"`:
  - Move such PRs to `Needs Maintainer Action` instead of `... Review`.
  - When producing the checklist comments, check `info.dangerLevel ===
    "Infrastructure"` earlier, to show that comment rather than mumble
    about approvals etc (since they're irrelevant: infra changes are
    almost always package removals).


elibarzilay:

PR #140, opened 2020-07-12
Minor dash to emdash

elibarzilay:
Missed it earlier.


elibarzilay:

PR #141, opened 2020-07-12
Remove some unused things, label a few that are kept

elibarzilay:
Keep a few that look like they might be useful in the future.

### DIFF
--- a/src/comments.ts
+++ b/src/comments.ts
@@ -40,17 +40,17 @@ export const ChangesRequest = (headOid: string, user: string) => ({
     status: `@${user} One or more reviewers has requested changes. Please address their comments. I'll be back once they sign off or you've pushed new commits or comments. If you disagree with the reviewer's comments, you can "dismiss" the review using GitHub's review UI. Thank you!`
 });
 
-export const ApprovedByOwner = () => ({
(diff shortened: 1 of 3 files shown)


elibarzilay:

src/comments.ts
```diff
@@ -40,17 +40,17 @@ export const ChangesRequest = (headOid: string, user: string) => ({
...
+export const AuthorIsOwnerAndGreen = () => ({ // UNUSED
     tag: "merge",
     status: "Since you're a listed owner and the build passed, this PR is fast-tracked. A maintainer will merge shortly. If it shouldn't be merged yet, please leave a comment saying so and we'll wait. Thank you for your contribution to DefinitelyTyped!"
 });
 
-export const LGTM = () => ({
+export const LGTM = () => ({ // UNUSED
     tag: "merge",
     status: "We've gotten sign-off from a reviewer 👏. A maintainer will soon review this PR and merge it if there are no issues. If it shouldn't be merged yet, please leave a comment saying so and we'll wait. Thank you for contributing to DefinitelyTyped!"
 });
```
orta, line 56: Yeah, they can all go since we merged all this sort of thing into one edited comment instead of a million bot comments

PR #142, opened 2020-07-12
Improve the situation for single owner packages

### LINKED ISSUE #105: Better workflow for "Author is Owner" / "No Other Owners"
Since recently, the demand for updating single owner definitions outstrips the capacity for DT maintainers to perform reviews. It used to be that I could count on adding features to my typings and it be no longer than 2 or 3 days and it would be merged but it's changed since the bot has taken over more responsibility. I am looking for a better workflow for single owner definitions. 

The single owner scenario should be treated with responsibility of maintaining the quality of the package, not a DT maintainer. It takes me about 8 hours to bring a certain definition up to date with a new JS release. A DT maintainer is not going to put as much effort into the review as I, the single owner, did and it's half insulting to assume so.

If the single owner is good enough for reviewing and allowing commits from others then they should surely be good enough for their own changes. They should be treated as if they own that part of the repo. In fact maybe having a way to allow git submodules to pull in definitions wouldn't be a bad idea (it would speed up the DT CI build). 

Looking at the current DT board. There are ~300 recently merged PRs...none of which are tagged "No Other Owners". Those are all waiting in other boards.

My current options are...
- Add another "owner" simply to allow the bot to ping someone else to review/approve the changes. Who is qualified? Who cares? The NPM Credits section shows "These definitions were written by" which would be technically incorrect unless it starts showing these definitions were Written By: Owners , Reviewed By: Everyone that's ever approved changes on the file.

- Wait for some unknown period of time until a volunteer comes along (~3 weeks?)

- Go try to hunt down a DT maintainer, and request their review. (Tried it and did not work)

- Become a DT maintainer? Does that allow me to approve my own changes?

- Forget using DT and maintain my own private repo.

elibarzilay:
* If this is a single-owner package, and its popularity is `Popular`,
  then make the required reviewers be "others" (and not "owners", which
  doesn't apply for these packages).

* In the case of single-owner packages whose popularity is not
  `Critical` (i.e., including the above case), spit out a comment
  encouraging the owner to find someone else to review and maybe
  recruit them as an owner.  (See my comment on #105.)

Closes #105.


elibarzilay:

elibarzilay:
If you think it's worth it, then I'll see if this can be re-done by changing `hasFinalApproval`...

PR #143, opened 2020-07-12
Handle errors, specifically failure to fetch file content

elibarzilay:
It looks like the GH api silently fails to fetch blobs that are around
5mb or bigger: it just returns an empty string in these cases.  However,
we can fetch the byteSize of the blob too, and using it detect when such
a failure happened.  (See below for a query that fails in this way.)

* Add `byteSize` to the file query, and use it to detect such failures
  and throw an error from `fetchFile`.

* Add `BotError` for cases where it's not a complete failure, but
  something that `process` can report in a comment.  (Clarify the roles
  of `BotError` and `BotFail` in comments.)

* Add an `error` field to `OwnerInfo` (move the `try`-`catch` block),
  and when there is one, use `botError` to return a value of the above
  type.

* Change a few `botFail`s to `botError`s for cases that can be similarly
  reported (no author, no head commit).  This leaves only the failure to
  get a PR object as a `BotFail`.

* Add a (very red) `Mergebot Error` label for these cases.

* For `BotError`s, use this label, move to the `Other` column, and post
  a comment with the error.

* In `daily.ts`, use `console.error` to show these errors so they're
  visible in the log too.

* (Also some minor linting, including indentation fixups.)

----

```
query GetLabels {
  repository(name:"DefinitelyTyped", owner:"DefinitelyTyped") {
    # e6fa521672 is where master is when I created this commit
    object(expression: "e6fa521672:types/arcgis-js-api/index.d.ts", ) {
      ... on Blob {
        __typename
        text
        byteSize
      }
    }
  }
}
```


elibarzilay:

PR #144, opened 2020-07-12
Update cache and improve updating

elibarzilay:
* Add a comment to the cache file saying how to update it.

* Make it search up for the repo directory, since otherwise it would
  sometimes produce the output in the wrong place.


elibarzilay:

PR #146, opened 2020-07-13
More work related to the trigger handler

elibarzilay:
* Make it handle `project_card` events with `moved` action (hacked from
  `project_card.content_url`).

* Rename `ForStatus` in all its forms to `ForSHA1` since that's what it
  takes, and some ObLint.

* Improved query to get the PR number: use `associatedPullRequests` from
  an `object` instead of a generic search
  ([mentioned here](https://developer.github.com/v4/changelog/2019-03-08-schema-changes/)).

  **No, scratch that...**  Looks like this fails *sometimes*, whereas
  the search work -- so use the old query, but keep the new one in a
  comment for reference when it works (or when there's a straightforward
  way to tell where the status came from).


elibarzilay:

orta:
Shame about the commit lookup, I did try using that the first time too but also had to back track to the search API

elibarzilay:
> Shame about the commit lookup, I did try using that the first time too but also had to back track to the search API

Heh, at least now it's in a comment, for the next person to try and make it better...

PR #147, opened 2020-07-13
Update markdown texts

elibarzilay:


elibarzilay:

PR #148, opened 2020-07-13
Major improvements for the daily cleanup and GQL queries

elibarzilay:
* Fix the deletion mutation, which was broken for about two months and
  cards were piling up instead of removed.

* Rename `recently-merged-prs-project` -> `projectboard-cards`,
  `getRecentlyUpdatedPRProjectBoardCards` -> `getProjectBoardCards`,
  etc.
  - Make it return *all* cards by column so the scan can remove closed
    PRs which were left over with cards.  (There were 3 of these,
    probably due to debugging.)
  - Make it return the total counts, so the daily run can report if
    there were additional cards that are left out of the deletion party.

* Rename `recently-updated-prs-query` -> `process-many-query` and leave
  a big FIXME note saying that it should be removed and implemented as a
  function around `all-open-prs-query`.  (They were both using the same
  type names, and the schema generator happily produced a file according
  to a random one from the two.)

* Change `getAllOpenPRs` into `getAllOpenPRsAndCardIDs`, so we get the
  cards that correspond with open PRs too.  (Rename this file too.)

* Rename `label-query.ts` -> `label-columns-queries.ts`.

There is a lot of noise due to all of the renaming, but I couldn't hold
back with the mess of everything being "RecentlyUpdated"...  The GH
projects page is reasonably responsive again for me, probably mainly due
to the actual cleanup of the recently merged column.


elibarzilay:

PR #151, opened 2020-07-13
Add `Edits Infrastructure` for `no_packages` PRs

elibarzilay:
Because it looks strange that there are no labels on these things.  Also
don't ignore them in `daily.ts`.


orta:
Yeah, good call

PR #154, opened 2020-07-17
(Try to) disable concurrent trigger runs

elibarzilay:
The `singleton` settings don't seem to have an effect, so try [this thing](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-http-webhook-output) instead.

Note that this limits all interactions — if the bots becomes busy enough, it would be better to limit concurrency per PR instead, somehow.


orta:
Looks good to me, let's give it a whirl!

PR #156, opened 2020-07-20
Allow owners to do a "ready to merge" self-merge

### LINKED ISSUE #98: There's no way to declare that module is auto generated
It's a follow-up to [discussion](https://discordapp.com/channels/508357248330760243/640177505915109377/723177888287031406) with @orta 
and it's related to https://github.com/Maxim-Mazurok/google-api-typings-generator/issues/245

I'm maintaining a [repo](https://github.com/Maxim-Mazurok/google-api-typings-generator) that is generating `d.ts` for Google client-side APIs (here's why: [Google API in NodeJS VS Client](https://github.com/Maxim-Mazurok/google-api-typings-generator#javascript-vs-nodejs-clients))

Google APIs are updated constantly. Almost every hour at least one API is changed. Sometimes it's just revision number bump. Sometimes it's change of wording in comments. Sometimes it's bigger.
So, we're using Github Actions in order to generate new typings and put them in `types` branch.
Another action compares our `types` branch with typings in DT and if it's different - opens/updates PR: [example PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45514).
You can find more info about this, including some flow-charts in [this PR](https://github.com/Maxim-Mazurok/google-api-typings-generator/pull/74) and linked issues.

There are currently a few problems:
1. When we're changing comments - bot still wants tests to be changed.
2. Another problem is when someone reviews my PR. [Good example](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45165). As you can see, bot kept moving it form one column one the project board to another, making it fo to the bottom of list every time.

Since we don't really care about revision-only updates, I [disabled them](https://github.com/Maxim-Mazurok/google-api-typings-generator/pull/151).

But still, it takes quiet some time to merge these changes and I'm feeling bad about wasting reviewers time.

Can we come up with some solution that would auto-merge auto-generated modules? Any ideas?

elibarzilay:
Add text to the comment telling owners that it can be done --- *except*
when there are too many owners, or when there are multiple packages.
(Owners can still do it if there are many, only multiple packages block
the whole thing.)

Closes #98.

Also remove an unused `const approvals` line.

### DIFF
--- a/src/_tests/fixtures/43144/mutations.json
+++ b/src/_tests/fixtures/43144/mutations.json
@@ -33,7 +33,7 @@
     "variables": {
       "input": {
         "subjectId": "MDExOlB1bGxSZXF1ZXN0Mzg4NDkxOTI2",
-        "body": "@jeffreymeng Everything looks good here. Great job! I am ready to merge this PR on your behalf.\n\nIf you'd like that to happen, please post a comment saying:\n\n> Ready to merge\n\nand I'll merge this PR almost instantly. Thanks for helping out! :heart:\n<!--typescript_bot_merge-offer-->"
+        "body": "@jeffreymeng Everything looks good here. Great job! I am ready to merge this PR on your behalf.\n\nIf you'd like that to happen, please post a comment saying:\n\n> Ready to merge\n\nand I'll merge this PR almost instantly. Thanks for helping out! :heart:\n\n(@pocesar: you can do this too.)\n<!--typescript_bot_merge-offer-->"
       }
     }
   }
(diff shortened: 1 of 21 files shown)


orta:
Cool, yeah - the code looks right to me, but I'm a bit unsure about why we get a lot more _new_ comments in the tests when the fixtures didn't get updated

elibarzilay:
There's a bunch of ignorable changes since there is one un-conditional new `\n` in the string now.  Could be removed with

    and I'll merge this PR almost instantly. Thanks for helping out! :heart:${
    otherOwners.length === 0 ? "" : `\n
    (${otherOwners.map(o => "@" + o).join(", ")}: you can do this too.)`}`});

But I suspected that it's too much.  (I'm fine doing this change it if you think it's worth it.)

PR #165, opened 2020-07-23
Temporary: Show failed data, disable sequential processing

elibarzilay:
Trying to debug why there is no `pullRequest` data in some cases.  Also,
disable sequential http processing to see if that causes the
weirdnesses.


orta:
👍  looks good to me

PR #166, opened 2020-07-23
Slight improvement for single owners

elibarzilay:
Make the main checklist show

> Most recent commit is approved by type definition owners or DT maintainers

It still is confusing that it's only a minor phrasing response for
blessing the PR, but it's still in the maintainer queue column.  It
might be better to have something more concrete for single owners, but
anything I can think about is a major change.

### DIFF
--- a/src/_tests/fixtures/46279/_downloads.json
+++ b/src/_tests/fixtures/46279/_downloads.json
@@ -1,0 +1,3 @@
+{
+  "phoenix_live_view": 1639
+}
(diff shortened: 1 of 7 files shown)


orta:
I don't think this works:

> `❌ Most recent commit is approved by type definition owners or DT maintainers`

But even then, switching the text here when it still shows as ❌ is probably more confusing to me?

elibarzilay:
The reason for the ❌ is the same as always -- it's a requirement that is not yet satisfied...

PR #169, opened 2020-07-28
Delay replies by 30s

### LINKED ISSUE #168: Delay on processing of new PRs or alternative processing
After I added information when failing to find a PR, I saw a failure that had the information below for a `pull_request` / `opened` event.  [The PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/46330) *does* exist, and its URL is even mentioned in the event payload.  So I'm guessing that there might be some processing delays (maybe there's a separate db that the gql api uses that needs to update) -- and therefore it's probably best to delay the processing of such events by a few seconds.

It's probably best to do this in addition to some manual per-pr sequentialization of handling incoming requests, since there are probably a small outburst of events when a pr is created.

Another possibility, since there are other cases when several events fire at the same time is, and since the processing always looks at the whole state so there's no need to process each and every event for the same PR:

* When an event is received, figure out the PR number, and *immediately* return some "OK" answer.
* Wait for several seconds for each of these (*after* returning the OK).
* In this period, if there are new events for the same PR, then ignore them (or maybe extend the waiting period for the existing one).
* Then do one processing of the PR.

```
Failed because of: No PR with this number exists, ({
  "data":{"repository":{"pullRequest":null,"__typename":"Repository"}},
  "loading":false,
  "networkStatus":7,
  "stale":false,
  "errors":[{
    "type":"NOT_FOUND",
    "path":["repository","pullRequest"],
    "locations":[{"line":3,"column":5}],
    "message":"Could not resolve to a PullRequest with the number of 46330."
  }]
})
```

elibarzilay:
And if another hook comes in for the same PR, then just do nothing, so
the later request will do everything.

Fixes #168

### DIFF
--- a/PR-Trigger/index.js
+++ b/PR-Trigger/index.js
@@ -8,6 +8,8 @@ const verify = require("@octokit/webhooks/verify");
 const sign = require("@octokit/webhooks/sign");
 const { runQueryToGetPRMetadataForSHA1 } = require("../bin/queries/SHA1-to-PR-query");
 
+const prHandlers = new Map();
+
 /** @type {import("@azure/functions").AzureFunction} */
 const httpTrigger = async function (context, _req) {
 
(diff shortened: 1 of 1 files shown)


elibarzilay:
Yeah, that's what I meant when I said that there's most likely no need for anything heavier like azure storage or whatever.  It *could* still do things in parallel if the azure function framework decides to have two instances or more running, but after looking at it for a while, it doesn't do that...

PR #170, opened 2020-07-28
Set author-attention labels with the rest

elibarzilay:
I don't think that it should make a difference since the default label
value is `false`, but maybe some mid-function failure made it sometimes
baheve as it did on DefinitelyTyped/DefinitelyTyped#44857.


elibarzilay:
Merging it because I want to track the bot behavior.

PR #171, opened 2020-07-28
A quick stab at improving the state of editing multiple packages

orta:
This attacks two problems:

 - You edit a dependency, then you have to update downstream tests as well as your own
 -  You see a non-sensical error message when blocked by being on multiple packages: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/46321

Will commenting inline

### DIFF
--- a/src/compute-pr-actions.ts
+++ b/src/compute-pr-actions.ts
@@ -407,6 +407,8 @@ function createWelcomeComment(info: PrInfo, staleness: Staleness) {
         const infraFiles = info.files.filter(f => f.kind === "infrastructure");
         const links = infraFiles.map(f => `[\`${f.path}\`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/${info.headCommitOid}/${f.path})`);
         display(` * ${emoji(approval.approved)} A DT maintainer needs to approve changes which affect DT infrastructure (${links.join(", ")})`);
+    } else if (info.dangerLevel === "MultiplePackagesEdited") {
+        display(` * ${emoji(approval.approved)} A DT maintainer needs to approve changes which affect more than one package`);
     } else if (info.dangerLevel === "ScopedAndTested" || info.maintainerBlessed) {
         display(` * ${emoji(approval.approved)} Most recent commit is approved by ${approval.requiredApprovalBy}`);
     } else if (otherOwners.length === 0) {
--- a/src/pr-info.ts
+++ b/src/pr-info.ts
@@ -421,6 +421,10 @@ export function getPackagesTouched(files: readonly FileInfo[]) {
     return [...new Set(noNulls(files.map(f => "package" in f ? f.package : null)))];
 }
 
+export function getPackagesWithSourceFilesTouched(files: readonly FileInfo[]) {
+    return [...new Set(noNulls(files.map(f => "package" in f && f.kind === "definition" ? f.package : null)))];
+}
+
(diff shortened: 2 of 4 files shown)


src/compute-pr-actions.ts
```diff
@@ -407,6 +407,8 @@ function createWelcomeComment(info: PrInfo, staleness: Staleness) {
         const infraFiles = info.files.filter(f => f.kind === "infrastructure");
         const links = infraFiles.map(f => `[\`${f.path}\`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/${info.headCommitOid}/${f.path})`);
         display(` * ${emoji(approval.approved)} A DT maintainer needs to approve changes which affect DT infrastructure (${links.join(", ")})`);
+    } else if (info.dangerLevel === "MultiplePackagesEdited") {
+        display(` * ${emoji(approval.approved)} A DT maintainer needs to approve changes which affect more than one package`);
```
orta, line 411: This used to fall through to the final else which is "without tests" - https://github.com/DefinitelyTyped/DefinitelyTyped/pull/46321 definitely has tests

src/pr-info.ts
```diff
@@ -421,6 +421,10 @@ export function getPackagesTouched(files: readonly FileInfo[]) {
     return [...new Set(noNulls(files.map(f => "package" in f ? f.package : null)))];
 }
 
+export function getPackagesWithSourceFilesTouched(files: readonly FileInfo[]) {
+    return [...new Set(noNulls(files.map(f => "package" in f && f.kind === "definition" ? f.package : null)))];
+}
```
orta, line 426: Felt pretty reasonable to do, basically the other 

````
type FileKind = "test" | "definition" | "markdown" | "package-meta" | "package-meta-ok"| "infrastructure";

````

is the full list, maybe it could also take into account `package-meta-ok`?

Comment (elibarzilay) [reply src/pr-info.ts:426]: It was originally "more typed", maybe it's best to return to some more structured thing, like with interface extensions, so there's a type for everything that has `package` and `suspect` would similarly exist only on `...-meta-ok`.

(It's a small thing though, so I hesitated to make it much bigger just for this.)

File: src/pr-info.ts [hunk 3]
```diff
@@ -498,9 +502,11 @@ function getDangerLevel(categorizedFiles: readonly FileInfo[]): DangerLevel {
         return "Infrastructure";
     }
     const packagesTouched = getPackagesTouched(categorizedFiles);
+    const sourcePackagesTouched = getPackagesWithSourceFilesTouched(categorizedFiles);
+
     if (packagesTouched.length === 0) {
         return "Infrastructure";
-    } else if (packagesTouched.length > 1) {
+    } else if (sourcePackagesTouched.length > 1) {
```
elibarzilay, line 509: I completely agree that it would be best to do something better with the multiple packages case, but I worry that this makes it easy to miss things, since it will show the multiple label only sometimes.

What I was thinking of is separating things so that multiple-packages is one attribute, and config/no-tests would be another--?  And if the search is done for all files then (for example) you won't get the untested label if you edited a definition in one package and a test in another (but you will get the multiple packages of course).

But regardless, it looks like this change wouldn't affect that PR (DefinitelyTyped/DefinitelyTyped#46321) since it has changes to two definitions files in two packages..

src/compute-pr-actions.ts
elibarzilay: Yeah, that should be added, but see below.

PR #172, opened 2020-07-29
Count packages with non-test files, instead of with definitions

elibarzilay:
Should be less easy to miss, but still cover the case of a pr with tests
in a different package.


orta:
Cool, works for me 👍

PR #173, opened 2020-07-29
List reasons for meta files that are not meta-ok

elibarzilay:
(Note that many tests list "couldn't fetch contents" as the reason,
since these tests were added before the contents was fetched and
checked.)


orta:
Good answer 👍

PR #191, opened 2020-09-02
Adds a / to the end of the unpkg url

orta:
For example in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/47224#issuecomment-685442884

it links to https://unpkg.com/browse/node-ip2region@1.0.2 which 404s, https://unpkg.com/browse/node-ip2region@1.0.2/ does not 👍🏻


.github/workflows/upload.yml
```diff
@@ -19,7 +19,6 @@ jobs:
         npm run build --if-present
         npm run test --if-present
     - uses: Azure/functions-action@v1
-      id: fa
```
orta, line 22: This isn't needed, because no other steps use the id

orta:
I've updated the GitHub secrets with the new publishing profile contents, so hopefully it will succeed on this deploy https://github.com/DefinitelyTyped/dt-mergebot/actions/runs/236627856

.github/workflows/upload.yml
elibarzilay: Yeah, I was wondering if it can be dropped too...

elibarzilay:
This is very weird -- I definitely tested it and dropped the slash when I saw that it's working, but now it fails.

(I'll merge and re-run the daily now so it'll update open PRs)

PR #201, opened 2020-10-18
Unify the repeated code that checks convergence to a required form

elibarzilay:
Use a generic helper for this, and improve it by wrapping a try-catch
only around the needed parts that parse or compare user data.

### DIFF
--- a/src/pr-info.ts
+++ b/src/pr-info.ts
@@ -398,48 +398,16 @@ configSuspicious["OTHER_FILES.txt"] = contents =>
     // all path parts are not empty and not all dots
     : !contents.split(/\n/).every(line => line === "" || isRelativePath(line)) ? "bad path"
     : undefined;
-configSuspicious["package.json"] = (contents, oldText) => {
-    // either the required form or deletions that get closer to it
-    const requiredForm = { private: true };
-    function diffFromReq(text: string) {
-        const json = JSON.parse(text);
-        for (const k of ["dependencies", "types", "typesVersions"]) {
-            delete json[k];
-        }
-        return jsonDiff.compare(requiredForm, json);
-    }
-    try {
-        const newDiff = diffFromReq(contents);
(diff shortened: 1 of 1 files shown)


elibarzilay:
@jablko --- that's the generalization that I wanted to do, specifically, I really didn't like the too-broad try-catch since they can swallow proper bugs.  (In fact they did in my first version since it didn't check for ignored dotted keys that don't exist...)

jablko (approved):
Nice! I'm 100% in favor.

src/pr-info.ts
```diff
@@ -451,30 +419,36 @@ configSuspicious["tsconfig.json"] = (contents, oldText) => {
...
+// helper for json file testers: allow either a given "requiredForm", or any edits that get closer
+// to it, ignoring some keys (sub-values can be specified using dots).  The ignored properties are
+// in most cases checked elsewhere (dtslint), and in some cases they are irrelevant.
+function makeJsonCheckerFromCore(requiredForm: any, ignoredKeys: string[]) {
+    const diffFromReq = (text: string) => {
+        let json: any;
+        try { json = JSON.parse(text); } catch (e) { return "couldn't parse json"; }
+        // allow dotted keys in ignoredKeys
+        ignoredKeys.map(k => k.split(".")).forEach(keys =>
+            delete keys.slice(0, -1).reduce((a,b) => a?.[b], json)?.[keys[keys.length - 1]]);
```
jablko, line 439: Probably I'd do, but personal taste/background :slightly_smiling_face:
```suggestion
        ignoredKeys.forEach(k => {
            const keys = k.split(".");
            delete keys.slice(0, -1).reduce((a,b) => a?.[b], json)?.[keys[keys.length - 1]];
        });
```

elibarzilay: As a possibly lame compromise, I broke the line there...

Issue #202, opened 2020-10-22
Ask people if they are sure they want to sign up as maintainer


alloy:
It happens too often that people mistake the ‘by’ header for _just_ a list of people that contributed to a file, not realising it means they are signing up to be a maintainer and getting pinged for each issue/PR. E.g. https://github.com/DefinitelyTyped/DefinitelyTyped/pull/48868#discussion_r510079733

It is cumbersome to manually ask this in a review and also might stall a PR longer than necessary if a bot could respond much faster. Is this the right place for that?

elibarzilay:
Yes, this is the right place.

One thing that I didn't like is how there are sometimes PRs that are big enough and/or noisy enough so it's easy to miss the fact that there's a new owner.  But I didn't realize that it is also common enough to add yourself to the list thinking that this is some required signing thing for all contributors...

So to kill both birds, maybe add some code that will add a label and/or text when a PR has any changes to the owners list?

The other issue here is the mistake of using an email (similar to people who leave it as `.../me`).  So maybe I'll also add code to verify that the change (adding/removing) is always for the PR author.  Non matches would be reported in the comment (so authors can quickly fix them) and also require a maintainer eyeball (to allow dropping unresponsive people.

alloy:
> So to kill both birds, maybe add some code that will add a label and/or text when a PR has any changes to the owners list?

A text comment [ideally on the change] would definitely be helpful to the author 👍 

I in turn didn’t realise it was an issue that people reviewing types of many packages might miss, so a label sounds great for that indeed.

> The other issue here is the mistake of using an email (similar to people who leave it as `.../me`). So maybe I'll also add code to verify that the change (adding/removing) is always for the PR author. Non matches would be reported in the comment (so authors can quickly fix them) and also require a maintainer eyeball (to allow dropping unresponsive people.

👌

elibarzilay:
Given the replies in DefinitelyTyped/DefinitelyTyped#49034, I think that it would be good to add a pointer to the readme section about owners to that text.

elibarzilay:
So, after some thinking, I've made #221, which adds lists of added/removed owners in the welcome comment, and also adds an `Edits Owners` label to PRs that do so unless they're for new/removed packages.

It doesn't address the feature request here completely --- specifically, there's no "are you sure?" comment.  Adding such information should be easy, but I worry about making it too verbose to be useful.  So hopefully the current additions will improve it by making the owner edits much more visible to the author and to reviewers.

[...]

Issue #178, opened 2020-08-19
People should not be able to self-merge until maintainers of all affected packages have approved a change.


alloy:
For example, in [this PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/46473) a maintainer of the `react-native-material-ripple` package approved the changes, but the bulk of the changes were actually in `react-native`, which the author was able to self-merge before a `react-native` maintainer was able to review the changes.

alloy:
There’s now the concept of ‘critical packages’, so this is done afaik.

elibarzilay:
There is, but it bears no implications for accepting a PR -- and actually that's what I'm working on adding right now.

Even when I look at PRs, when there are multiple packages I'd wish there was a mapping of package to owners so I have an idea which packages are fine with the change.  So this will be useful for more than just making self-merges more robust.

PR #205, opened 2020-10-27
Major refactor in preparation of additional functionality

elibarzilay:
* Changes how information is stored:
  - `pr-info.ts`: create a new `pkgInfo: PackageInfo` field to hold
    per-package info, including `FileInfo[]`, instead of scattered
    generalized bits,
  - `compute-pr-actions.ts`: use the new layout instead,
  - `derived.json` files changed, but no functionality changes (no other
    test changes, but see below for some minor adjustments).

* Move `noNulls`, `notUndefined` to `util/util`.

* New `ExtendedPrInfo` to hold "second order derived info", instead of
  dragging around a bunch of values.  This includes an `orig` pointer to
  the original `PrInfo` which is what gets writted to the diagnostics
  part of the welcome message (might be more convenient to include it in
  the future?).

* Use this to remove a bunch of data that is under `pkgInfo` while still
  making it easy to access it.  (Note the exported `pkgInfoAllOwners`
  utility, used to avoid keeping duplicate owner information.)

* Use a `getMonthlyDownloadCount`-like function in
  `fixturedActions.test.ts`, so it is possible to abstract over one
  function, use a default value, and avoid the double implementation of
  the popularity function.

  - This require some small test fixes: a few missing `_downloads.json`
    files, and a bogus package name in the `38979` file.

* New information:

  - per-package popularity level, working towards #178.  (Note: not
    keeping the download numbers since those are likely to change very
    frequently.)

  - New information: `addedOwners` & `deletedOwners`, showing owner
    changes, working toward #202.  (Adjust `_files.json` files to get
    the correct owner changes.  Includes a change to the `46879` test:
    it didn't have any files, so it was incorrectly labeled as a new
    package.)


src/pr-info.ts
```diff
@@ -582,70 +603,21 @@ function getCIResult(headCommit: PR_repository_pullRequest_commits_nodes_commit)
...
-    if (arr == null) return [];
-    return arr.filter(arr => arr != null) as T[];
+    if (indexDtsContent === undefined) return null;
+    let parsed: HeaderParser.Header;
+    try {
+        parsed = HeaderParser.parseHeaderOrFail(indexDtsContent);
+    } catch (e) {
+        if (e instanceof Error) return new Error(`error parsing owners: ${e.message}`);
+    }
+    return parsed!.contributors.map(c => c.githubUsername).filter(notUndefined);
```
orta, line 622: I wonder if we _don't_ use the ./CODEOWNERS from DT in this codebase, looking at this.

orta (approved):
Nice one, yeah, looks like a solid architectural improvement

PR #211, opened 2020-10-30
Add engines to recommend node 12, and adds node12 to the upload yml

orta:
Should force contributors to have the minimum version, and allows us to deploy 👍🏻


orta:
checked azzurre, we're on runtime 3 - which is node 12 👍🏻 

Screen Shot 2020-10-30 at 8 10 25 AM

elibarzilay (approved):

PR #214, opened 2020-11-02
Revert to node v10 and minor shuffles

elibarzilay:
* Go back to node v10 because apparently you can't get v12 from azure
  functions.

* Add a `flatten` utility, and as long as I'm there, also `unique`.  Use
  them to avoid `.flat(1)`.

* Remove `src/util/dt.ts` which is an old leftover?


andrewbranch (approved):

jablko:
I think v12 is the default in Functions runtime 3.x? https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-node?tabs=v2#node-version

Do you happen to know which runtime we're currently on? https://docs.microsoft.com/en-us/azure/azure-functions/set-runtime-version?tabs=portal%2Cazurecli-linux

elibarzilay:
The logs show `v10.23.1`.

Issue #206, opened 2020-10-27
Improve "multiple packages" label?


elibarzilay:
Currently, the multiple packages label is set if `dangerLevel` is `"MultiplePackagesEdited"`, but this is only for `nonTestPackagesTouched` so IIUC, it doesn't look at packages with test.  Should it be improved to appear whenever there are multiple packages regardless of test?

orta:
Good question, one answer:

 - Check for the highest popularity package for tests 
    e.g. I improve the types for x, but smaller libs who depend on x get a one liner update with no tests can be OK'd
    this could be a set of package maybe if there are multiple

I wonder how much this is also accounted for with the multiple package sign-off work, like if I sign off on the most popular lib that should probably be ok for the smaller ones?

elibarzilay:
OK, so I take it that this behavior (putting the multiple-files label based on `dangerLevel == MultiplePackagesEdited`) is as intended, and I also noticed that this is the subject of #155.

And the further improvement that you're suggesting is to assign this danger level when there are multiple packages that are either untested *or* critical -- right?

orta:
Yes!

Issue #155, opened 2020-07-19
Don't take test files into account when looking at whether a package affects multiple packages


orta:
See https://github.com/DefinitelyTyped/DefinitelyTyped/pull/46112

Could also do with changing the 3rd statement in the main message for multiple packages on the whole too

PR #215, opened 2020-11-03
More info shuffling and refactoring

elibarzilay:
* The code for a `dangerLevel` of `MultiplePackagesEdited` was broken:
  it used `noNulls` over a list of booleans, which means that
  `MultiplePackagesEdited` corresponded with just having multiple
  packages (not taking into account tests or criticals).  This changes
  the code to use just a simple count of packages for now.

  I tried to fix it, but that exposes a problem with `dangerLevel` in
  general: it is sometimes used as an indicator for a single package
  literally (eg, allow all owners to request a merge if it starts with
  "Scoped") and sometimes as a generic danger situation.  This is likely
  my fault since pretty much any use of it confused me (since it's
  defined as a hard-to-remember combination of different things).

  For the same reason, I think that it'd be best to try to remove it and
  use the information directly.  Some of the following changes are going
  in this direction.  I will revisit the issues around it when this is
  done (eg #206 / #155).

* Use an extended `editsInfra` field instead of `dangerLevel ===
  "Infrastructure"`.

* Working toward the above, move the `getDangerLevel` computation (and
  the `dangerLevel` field, and the `DangerLevel` type) to
  `ExtendedPrInfo` since there are no more uses of it in `pr-info`.  The
  next step is to split it into fields that express the relevant parts
  instead of a single level value.

* Move `authorIsOwner` from a `PrInfo` field to a computed
  `ExtendedPrInfo` one.  It's meaning is also changed: previously, it
  would be on if the author is an owner of *any* package, and now it's
  on if the author is an owner of *all* packages (if more than one).
  This currently affects only the corresponding label.  Two
  tests (45836, 45946) updated for this.

* Aggregate all review info into a single `getReviews` (much revised
  from `analyzeReviews`) that returns a `review: ReviewInfo[]`, holding
  all of the relevant information.  Remove all of the other fields, and
  add the ones that are needed into the `ExtendedPrInfo`.

  There are some minor diffeences --- for example, previously all stale
  reviews were kept, and now it's only one review kind per person so at
  most one stale review.  These differences only affect the derived
  information, but no test changes.

* Note that the computation of `approvalFlags` moved to `extendPrInfo`
  (in `compute-pr-actions`), where it can be changed to adapt it to
  address #178.

### DIFF
--- a/src/_tests/fixtures/38979/derived.json
+++ b/src/_tests/fixtures/38979/derived.json
@@ -3,7 +3,6 @@
   "now": "2020-05-12T19:14:42.000Z",
   "pr_number": 38979,
   "author": "ExE-Boss",
-  "dangerLevel": "ScopedAndConfiguration",
   "headCommitAbbrOid": "2223341",
   "headCommitOid": "222334139e52fc16369464cfb5dc95c82f71192f",
   "mergeIsRequested": false,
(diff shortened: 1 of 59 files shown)


orta (approved):
Nice work, looks like net lines removed 👍🏻

PR #219, opened 2020-11-09
Get monthly download count for the month up to the `lastPushDate`

### LINKED ISSUE #209: Avoid download count wobble
We're currently using the download count for "last-month", which can lead to the popularity level wobbling (eg, DefinitelyTyped/DefinitelyTyped#46778).  Instead, would be better to use a specific range of last pr update down to a month before that ([description here](https://github.com/npm/registry/blob/master/docs/download-counts.md)).

elibarzilay:
Actually, use the month up to a week before `lastPushDate`, in case it
takes npm some time to update the numbers.

This should make the download numbers stable, avoiding the wobble seen
in DefinitelyTyped/DefinitelyTyped#46778.

Fixes #209.


orta:
Clever 👍🏻

PR #221, opened 2020-11-10
Add information on owner changes, and more

### LINKED ISSUE #202: Ask people if they are sure they want to sign up as maintainer

elibarzilay:
* In the collected package info, add an explicit `kind` field specifying
  whether a package change was an `add`, `delete`, or `edit`.  This
  replaces the previous thing of having `owners === null` iff it's a new
  package.

* Add a new `Edits Owners` label, turned on when there are owner
  addition or deletions ignoring added/removed packages.

* Note that `delete` is really an approximation since it's the value
  used when `index.d.ts` is missing from the head commit.  A future
  extension should test that the file is deleted if-and-only-if all
  package files are deleted.  (And similarly, throw an error if a new
  package has a missing `index.d.ts`.)

* In the list of packages, show for each the list of added / removed
  owners.  Use a "✎" next to the author so it's easy to see changes done
  to other owners.

* Also add an "are you sure you want to become and owner" comment for
  PRs that add the author as an owner to a package with 4 or more
  owners, and the author is not an owner of another package in the same
  PR.

* Sort out some confusion with empty lines between secions.

* Remove setting `reviewerAdvisory` first, and instead move the code to
  just `display`ing the lines where the former was used.

* Added `isAuthor` predicate in the `ExtendedPrInfo` which is a
  specialization of `sameUser`.  (This is the first function in the
  info.)

Closes #202.


src/compute-pr-actions.ts
```diff
@@ -3,6 +3,7 @@ import { PrInfo, BotError, BotEnsureRemovedFromProject, BotNoPackages } from "./
 import { CIResult } from "./util/CIResult";
 import { ReviewInfo } from "./pr-info";
 import { noNulls, flatten, unique, sameUser } from "./util/util";
+import { userInfo } from "os";
```
orta, line 6: Likely a mistake

src/compute-pr-actions.ts
```diff
@@ -122,6 +126,7 @@ interface ExtendedPrInfo extends PrInfo {
     readonly otherOwners: string[];
     readonly noOtherOwners: boolean;
     readonly tooManyOwners: boolean;
+    readonly editsOwners: boolean;
     readonly canBeMerged: boolean;
```
orta, line 130: I know this is out of scope, but it just came up in discord: this should probably be renamed to `canBeSelfMerged` or `autoMerged` / `canBeOwnerMerged` or something like that

orta (approved):
Cool, looks good

src/compute-pr-actions.ts
elibarzilay: Yeah, I caught that in one of the upcoming PRs, so I'll leave this pr as is for now...

(An annoying feature of the Emacs ts thing is that if I mistype and complete a random binding, it will add the import...)

src/compute-pr-actions.ts
elibarzilay: OK, I'll do that in the next PR.

PR #226, opened 2020-11-12
`compute-pr-actions`: reorganize code

elibarzilay:
The idea is to keep all extra information computation in `extendPrInfo`,
with anything that is not a relatively simple one-liner being pushed to
the end of the function.

* Move all types to the top.

* Move all perperty helpers into `extendPrInfo` and simplify them.

* Rename `projectBoardForReviewWithLeastAccess` -> `reviewColumn`, and
  move the corresponding function into `extendPrInfo` too.

* Create a new `getApprovalFlags` function instead of an inlined
  expression.

* Fix a small bug: `getApprovalFlags` now returns `ApprovalFlags.None`
  if `hasChangereqs`.  (Test added: 48945)

* Another bug, due to GH: sometimes the API returns null for the
  `pushedDate` of a commit.  Use the PR `createdAt` in these
  cases.  (Not `committedDate`/`authoredDate`, since these can be set to
  arbitrary values.)  (Test added: 48708)

* Drop a couple of unused imports.


src/compute-pr-actions.ts
```diff
@@ -167,23 +178,73 @@ function extendPrInfo(info: PrInfo): ExtendedPrInfo {
...
+    const reviewColumn = getReviewColumn();
     return {
         ...info, orig: info, reviewLink,
         authorIsOwner, editsInfra, editsConfig, allOwners, otherOwners, noOtherOwners, tooManyOwners, editsOwners,
         canBeMerged, approved, approverKind, requireMaintainer, blessable, failedCI, staleness,
         packages, hasMultiplePackages, hasTests, newPackages, hasNewPackages,
         approvedReviews, changereqReviews, staleReviews, approvalFlags, hasChangereqs,
-        isAuthor
+        reviewColumn, isAuthor
     };
```
orta, line 196: yeah, this is very typescript compiler-y

orta (approved):

PR #230, opened 2020-11-15
Require approvals of *all* critical package owners

### LINKED ISSUE #178: People should not be able to self-merge until maintainers of all affected packages have approved a change.

elibarzilay:
Note that this is relevant only when a PR is blessed (which should
become a frequent tool), since critical packages means that the PR
starts on the maintainer queue, and if the maintainer approves then no
owner approvals are needed.  When a PR is in this state (multiple
critical packages and blessed): show a list of approvals done and needed
for each of the critical packages.

This also changes the "Owner Approved" label meaning to mean "All
Critical Package Owners" when there is more than one critical package.

And also show the list of critical packages whose owners still need to
approve the PR.  (This is shown regardless of blessing, so maintainers
can get the picture when considering whether to bless or approve or
whatever.)

Closes #178.

Also:

* Get rid of `requiredApprovers` and use a single form for all mentions,
  the text reads better in all of the changed tests.

* rename `canBeMerged` -> `canBeSelfMerged`.


orta (approved):
Cool, looks good

PR #232, opened 2020-11-19
Rework self merges

### LINKED ISSUE #216: Improve search for a relevant "ready to merge"
The problem is that using `firstApprovalDate` can kick in when the first approval is irrelevant for Merge:Auto, so what can happen, as demonstrated in DefinitelyTyped/DefinitelyTyped#48652 is:

* A review needs a maintainer approval
* It gets a random approval
* Author says "ready to merge" -- which is ignored since it's not in Merge:Auto
* Maintainer approves
* The bot moves it to Merge:Auto,
  - notices the author request which is *after* the random approval
  - that approval is the first
  - therefore the bot immediately merges the PR

To solve this:

1. Use `lastApprovalDate` instead, so the above scenario would ignore the request.
2. If there is *any* "ready to merge" requests from author/owner/s A,B,C, which are before the `sinceDate` argument (which uses the above), post a JFYI for A,B,C (with comment that the others can do so too) that the earlier request is ignored due to new activity, and it should be re-posted.
3. Actually, revise (2): instead of all R2M requests before `sinceDate`, look for the requests that are before that date *and* after the last "JFYI" note.  This would avoid repetitions.

The above scenario should be "fixed" even if:
* The bot is dead for some reason when the maintainer approves it, followed by a R2M request, followed by someone random approving it: it would still require a new R2M request (technically no need since it's the new *random* review that made it needed).  In this case it would be a bit weird anyway, but the author would know to post a new request.
* Instead of a review, the maintainer blesses the PR.  This means that the random review *is* the one that made it into Auto:Merge, and since the R2M comes after that it would kick in and it'll be merged immediately.

elibarzilay:
## Rework self-merges wrt premature requests

* Add the head sha1 to the self-merge offer comment (which GH should
  turn to a commit link).

* New tests:
  - `48652-merge-offer`: test making the offer comment
  - `48652-retract-merge-offer`: test deleting the offer comment on
    changereq

* Move `parseComment`/`makeComment` to a new `util/comment.ts`, so
  `parse` can be used to find a merge-offer in `pr-info`.

* Get the most recent merge offer date in `mergeOfferDate`.

* Add `mergeRequestDate`, `mergeRequestUser`, instead of
  `mergeIsRequested` which is now added in `extendPrInfo`.

(Motivated by and supersedes #227 by @jablko.  Originally used some of
the code from it, but due to the different approach and additional work
that was merged ended up being quite different.)

Fixes #216

## Add "please wait for an offer" comments for premature ready-to-merge

Requires the actual merge offer comment to appear, and if it isn't found
then post a message telling the requestor to wait until it does.

Since it relies on the comment, there is a very short timeframe (while
the bot is processing new events, around a minute) where the request is
invalid only because the offer wasn't posted.  The idea is to make
people get used to the idea that they cannot expedite the process by
pre-requesting a merge (which is exactly what leads to instant merge
surprises that nobody likes).

Part of the fix for #216.

## Rename `Merge:*` labels

These are relics of the previous workflow, and they're no longer good.

* Rename "Merge:YSYL" -> "YSYL"
  With the upcoming timeline changes "YSYL" will be applied
  when *auto-closing* a PR.

* Rename "Merge:Auto" -> "Self Merge"
  Since this is actually doing a merge-on-request rather than anything automatic.


src/pr-info.ts
```diff
@@ -427,12 +435,26 @@ function makeJsonCheckerFromCore(requiredForm: any, ignoredKeys: string[]) {
...
-        && (new Date(comment.createdAt)) > sinceDate);
+function latestComment(comments: PR_repository_pullRequest_comments_nodes[]) {
+    if (comments.length === 0) return undefined;
+    return comments.reduce((r, c) => r && Date.parse(r.createdAt) > Date.parse(c.createdAt) ? r : c);
+}
+
+function getMergeOfferDate(comments: PR_repository_pullRequest_comments_nodes[], abbrOid: string) {
+    const offer = latestComment(comments.filter(comment =>
+        sameUser("typescript-bot", comment.author?.login || "-")
+        && Comment.parse(comment.body)?.tag === "merge-offer"
```
orta, line 446: This capitalization really threw me off for a second, not offensive but was a tad suspicious

src/pr-info.ts
```diff
@@ -226,13 +238,10 @@ export async function deriveStateForPR(
...
         headCommitAbbrOid: headCommit.abbreviatedOid,
         headCommitOid: headCommit.oid,
-        mergeIsRequested: !!prInfo.comments.nodes
-            && usersSayReadyToMerge(noNulls(prInfo.comments.nodes),
-                                    pkgInfo.length === 1 ? [author, ...pkgInfo[0].owners] : [author],
-                                    latestDate(createdDate, lastPushDate, reopenedDate, firstApprovalDate) || lastPushDate),
         stalenessInDays,
         lastPushDate, reopenedDate, lastCommentDate,
         maintainerBlessed: lastBlessing ? lastBlessing > lastPushDate : false,
+        mergeOfferDate, mergeRequestDate: mergeRequest?.date, mergeRequestUser: mergeRequest?.user,
```
orta, line 244: nice move

src/comments.ts
```diff
@@ -81,6 +82,14 @@ and I'll merge this PR almost instantly. Thanks for helping out! :heart:
 ${otherOwners.length === 0 ? "" : `
 (${otherOwners.map(o => "@" + o).join(", ")}: you can do this too.)`}`});
 
+export const WaitUntilMergeIsOK = (user: string, abbrOid: string) => ({
+    // at most one reminder per update
+    tag: `wait-for-merge-offer-${abbrOid}`,
+    status: `Hi @${user}, I can't accept a merge request until the PR has a green CI and was appropriately reviewed. I will let you know once that happens.
```
orta, line 88: Yeah, I like this wording 👍🏻

src/comments.ts
```diff
@@ -81,6 +82,14 @@ and I'll merge this PR almost instantly. Thanks for helping out! :heart:
...
+
+Thanks, and happy typing!`
```
orta, line 90: and this is cute.

Maybe we _could_ link to [the overview](https://github.com/DefinitelyTyped/dt-mergebot/blob/master/docs/dt-mergebot-lifecycle.png)

src/_tests/fixtures/48652-retract-merge-offer-and-prerequest/mutations.json
```diff
@@ -0,0 +1,71 @@
...
+  },
+  {
+    "query": "mutation($input: AddCommentInput!) { addComment(input: $input) { clientMutationId } }",
+    "variables": {
+      "input": {
+        "subjectId": "MDExOlB1bGxSZXF1ZXN0NTAwNjg4ODM0",
+        "body": "Hi @mgol, I can't accept a merge request until the PR has a green CI and was appropriately reviewed. I will let you know once that happens.\n\nThanks, and happy typing!\n<!--typescript_bot_wait-for-merge-offer-06d8d67-->"
+      }
+    }
+  },
```
orta, line 62: 👍🏻

orta (approved):
Cool looks good to me, nice thinking

src/comments.ts
elibarzilay: Yeah, slightly tweaked Ryan's version since there are other reasons besides review that can make it wait for a merge offer...

src/pr-info.ts
elibarzilay: Which one?  I don't mind changing...

(`Comment` is similar to `Comments`, if that's the one...)

src/comments.ts
elibarzilay: Good idea, I made it point [here](https://github.com/DefinitelyTyped/DefinitelyTyped#make-a-pull-request) instead of directly at the image.

src/pr-info.ts
orta: `Comment` I thought it was a `class`

src/util/util.ts
```diff
@@ -52,10 +52,10 @@ export function forEachReverse<T, U>(arr: readonly T[] | null | undefined, actio
 }
 
 export function earliestDate(...dates: (Date | undefined)[]) {
-    return dates.reduce((d1,d2) => (d1 && d2 && d2 < d1 ? d2 : d1) || d2, undefined);
+    return dates.reduce((d1,d2) => (d1 && d2 && d2 < d1 ? d2 : d1) || d1 || d2, undefined);
```
jablko, line 55: Just curious about these, why the change?

elibarzilay: Bah.  It wasn't supposed to be there.

(I debated for a while whether `(...) || d2` is being too much of a smartass and thought that maybe it's better to add `d1` just for the sake of clarity.  Then I decided against it, but forgot to remove it.  Now I think that the `d1`/`d2` names are confusing since they're not symmetrical, so I'll do that instead.)

src/pr-info.ts
elibarzilay: OK, I'll change it.

PR #241, opened 2020-11-26
Revise staleness timelines and timeline handling

### LINKED ISSUE #204: Problem with long time to mark a PR abandoned
This is a result of having a month for getting a PR into abandoned state: when a PR was in "waiting for reviews" for a month and got to Abandoned state and therefore on the dt queue --- in this state I hesitate to ping the author or owners since my comment would restart the timer and no maintainers would see it for another month.

I've seen PRs where a maintainer pinged people, and much time passes after that, or where the author tries to ping the owners every 2-3 weeks which leads to PRs that linger for a very long time.

Possible solutions:

* Ignore certain comments (eg, ignore comments from dt maintainers, or comments that contain "ping").  But what about a reply to a ping?  -- It shouldn't reset the counter too.

* Count staleness (either when in a good state or with a failed build) from the last PR update instead of from the last activity.  (But this is a bit ugly since it should only apply for the long timelines.)

* Similar alternative: make the staleness clock go for the short week time if we're a month away from the last pr update -- something like `max(lastActivity + 7days, lastUpdate + 1month)`.  (Where `7days` and `1month` are actually the short and long periods.)

The last one looks best to me.

### LINKED ISSUE #207: "Abandoned" is a confusing/bad label for PRs that don't fail
Currently, the bot has three categories for "expired" PRs:

* YSYL: the pr is ready to merge but wasn't.
* Abandoned + failing: the PR is auto-closed.
* Abandoned + passing: happens due to lack of reviews.

(The first and last make the PR go to the maintainer NMA queue.)

Should use a better label for the last one, since it wasn't abandoned.  Change that to `Unreviewed`.

### LINKED ISSUE #208: Suggested timeline changes
### YSYL timeline:

(Currently *one week => NMA*; changes to *one month => closed*.)

- Measure time from when the PR *became mergable* (*excluding* comments etc)
- Wait a week
- Post a ping comment for author (+ owners) with a "JFYI will be closed in 3 weeks if not merged"
- Wait three more weeks (so the total time changes from a week to a *month*)
- not merged yet: apply YSYL label and *close it*
  (this is a change from the previous behavior of putting it on the maintainer NMA queue)

### Unreviewed timeline:

(Currently *one month => NMA*; changes to *two weeks => NMA*.)

- Measured from the *creation time* (including a draft->ready change, or a closed-and-reopened)
- The PR is on either waiting-for-reviews or needs-maintainer-review if there's something to pay attention to
  (in the can be blessed away from NMR to WFR)
- Wait *10 days*
- Post an "unreviewed for over a week" re-ping comment for reviewers
  (do not mention that it will go to NMA, to avoid laziness)
- Wait *1 week*
- Put on NMA (where a maintainer needs to do something)

elibarzilay:
* `staleness` is now an optional object with fields for the
  `state` (fresh / attention / nearly / done) `kind` (which is also the
  label that will be applied when the timeline is done), and `days` for
  the count (replacing `stalenessInDays`).  It's `undefined` when we're
  not on any of the timelines.

* Shuffle the code so staleness handling is used in the same place
  regardless of other state.  Note that it's done regardless of other
  actions which leads to some changes test in tests, but in practice it
  shouldn't be a problem (unless the bot dies for many days).

* The `Staleness` generation becomes the place for configuring staleness
  timelines, together with mappings in `comments.ts` for the explanation
  and comments of various states.

* Added the `Unreviewed` kind --- changed the 44857 to use it instead of
  `Abandoned`.

* Staleness also has a `doTimelineActions` which is now the central
  place for doing all timeline-related actions.

* Change: the stalness comment tags are computed (in `makeStaleness`),
  they depend on the state:
  - `nearly`: a tag that includes the date that the countdown started
    at, so these comments can be repeated if there's a timeline change
    that voided the previous staleness, but a new one triggered.
  - `done`: no date in the tag, since these comments are terminal.

* Change: put the `YSYL`/`Abandoned`/`Unreviewd` label when reaching
  `nearly` state, so it is useful to find PRs that are approaching their
  terminal (`done`) state.

* Revise the timelines, with some minor changes, and update all comments
  accordingly.  Notable changes from #208:
  - The waits are slightly different, but roughly the same.
  - "Becoming mergable" is measured from the merge offer comment when
    found.
  - The unreviewed timeline counts days from the last update rather than
    from the creation time.  This because I see a lot of PR's that have
    a few days of work following their creation.  It could be confusing
    if the count starts at creation, since some of these PRs would seem
    to be announced unreviewed very quickly.

Closes #208
Closes #207
Closes #204


src/_tests/fixtures/38979/mutations.json
```diff
@@ -56,5 +57,14 @@
...
     }
+  },
+  {
+    "query": "mutation($input: AddCommentInput!) { addComment(input: $input) { clientMutationId } }",
+    "variables": {
+      "input": {
+        "subjectId": "MDExOlB1bGxSZXF1ZXN0MzI1ODk5Njc0",
+        "body": "It has been more than two weeks and this PR still has no reviews.\n\nI'll bump it to the DT maintainer queue. Thank you for your patience, @ExE-Boss.\n\n(Ping @RReverser.)\n<!--typescript_bot_Unreviewed:done-->"
+      }
+    }
```
orta, line 68: Woo! 🌮

src/compute-pr-actions.ts
```diff
@@ -164,30 +168,29 @@ function extendPrInfo(info: PrInfo): ExtendedPrInfo {
...
-               : stalenessInDays <= 22 ? Staleness.PayAttention
-               : stalenessInDays <= 30 ? Staleness.NearlyAbandoned
-               : Staleness.Abandoned);
+        const mkStaleness = makeStaleness(info.now, info.author, otherOwners);
+        if (canBeSelfMerged && info.mergeOfferDate) return mkStaleness(
+            "YSYL", info.mergeOfferDate, 4, 9, 30, "CLOSE");
+        if (needsAuthorAction) return mkStaleness(
+            "Abandoned", info.lastActivityDate, 6, 22, 30, "CLOSE");
+        if (!approved) return mkStaleness(
+            "Unreviewed", info.lastPushDate, 6, 10, 17, "Needs Maintainer Action");
```
orta, line 192: Cool, this is clearer 👍🏻

orta (approved):
Nice work on centralizing the logic, all looks pretty solid to me 👍🏻

src/compute-pr-actions.ts
```diff
@@ -394,6 +379,35 @@ export function process(prInfo: PrInfo | BotEnsureRemovedFromProject | BotNoPack
...
+        const days = daysSince(since, now);
+        const state = days <= freshDays ? "fresh" : days <= attnDays ? "attention" : days <= nearDays ? "nearly" : "done";
+        const kindAndState = `${kind}:${state}`;
+        const explanation = Comments.StalenessExplanations[kindAndState];
+                const comment = Comments.StalenessComment(author, otherOwners)[kindAndState];
+        const doTimelineActions = (context: Actions) => {
+            if (comment !== undefined) {
+                const tag = state === "done" ? kindAndState
+                    : `${kindAndState}:${since.toISOString().replace(/T.*$/, "")}`;
+                const status = comment.replace(/\$\{user\}/, author);
```
jablko, line 395: Help me see this, it looks like `StalenessComment` are all template literals?

src/compute-pr-actions.ts
```diff
@@ -394,6 +379,35 @@ export function process(prInfo: PrInfo | BotEnsureRemovedFromProject | BotNoPack
...
+            doneColumn: ColumnName | "CLOSE") => {
+        const days = daysSince(since, now);
+        const state = days <= freshDays ? "fresh" : days <= attnDays ? "attention" : days <= nearDays ? "nearly" : "done";
+        const kindAndState = `${kind}:${state}`;
+        const explanation = Comments.StalenessExplanations[kindAndState];
+                const comment = Comments.StalenessComment(author, otherOwners)[kindAndState];
+        const doTimelineActions = (context: Actions) => {
+            if (comment !== undefined) {
+                const tag = state === "done" ? kindAndState
+                    : `${kindAndState}:${since.toISOString().replace(/T.*$/, "")}`;
```
jablko, line 394: Am I right that the date in the tag isn't consumed yet?

elibarzilay: The tag *contains* the date.  The reasoning is that when a staleness timeline starts, the tag for the comment would be based on the day it started, which means that if a PR gets out of the staleness timeline and later gets back into it, there would be a date for that, and therefore a second post.

src/compute-pr-actions.ts
elibarzilay: Yeah, note the comment there:

> Comments to post for the staleness timeline (the tag is computed in `makeStaleness`)

The reason for that is that the tag calculation is easier to do here, making it depend on the staleness when it's declared stale, or on that together with the date when it's the nearly stale comment.

jablko: Sorry, where's the `${user}` placeholder? I should pull and grep the changes, I haven't found it via the GitHub UI.

src/compute-pr-actions.ts
jablko: Sounds good, where does that happen?

src/compute-pr-actions.ts
elibarzilay: Oh, *that*!  You're right, it's a bad leftover that I should fix, thanks!

(I started with a hack of regexp-replacing `${user}` in the string and then switched to a normal argument w/ a template string, and therefore this bit is redundant.)

jablko: My bad vague! All good!

src/compute-pr-actions.ts
elibarzilay: Either I'm not following you or the bogus replacement line was confusing...?  (It's used in the comment that is pushed right after this bit.)

jablko: I'm searching for the code that pulls the date out of an existing comment and uses it for the described behavior.

elibarzilay: The date string is in the `tag`, `getMutationsForComments` (in `execute-pr-actions`) uses `comment.make` to embed the tag as an html comment in the posted comment, and it also searches for an existing comment with the same tag to decide whether it should be a new comment, or editing an existing one, or nothing at all.

jablko: Ohhh sorry it's an identifier to discriminate comments originating from different countdowns (re-reading your explanation and considering the purpose of `tag` it's obvious). Thanks!

jablko (approved):

src/compute-pr-actions.ts
elibarzilay: Yes, exactly.

src/compute-pr-actions.ts
```diff
@@ -88,13 +94,12 @@ const uriForTestingNewPackages = `${baseURI}#testing`;
...
-    NearlyYSYL,
-    YSYL,
-    NearlyAbandoned,
-    Abandoned,
+type Staleness = {
+    readonly kind: StalenessKind;
+    readonly days: number;
+    readonly state: "fresh" | "attention" | "nearly" | "done";
+    readonly explanation?: string;
+    readonly doTimelineActions: (context: Actions) => unknown;
```
jablko, line 102: Just curious: I'd've expected `void`?
```suggestion
    readonly doTimelineActions: (context: Actions) => void;

Comment (elibarzilay) [reply src/compute-pr-actions.ts:102]: Yeah, `void` is better.

File: src/compute-pr-actions.ts [hunk 6]
```diff
@@ -394,6 +379,34 @@ export function process(prInfo: PrInfo | BotEnsureRemovedFromProject | BotNoPack
     return context;
 }
 
+function makeStaleness(now: string, author: string, otherOwners: string[]) { // curried for convenience
+    return (kind: StalenessKind, since: Date,
+            freshDays: number, attnDays: number, nearDays: number,
+            doneColumn: ColumnName | "CLOSE") => {
+        const days = daysSince(since, now);
+        const state = days <= freshDays ? "fresh" : days <= attnDays ? "attention" : days <= nearDays ? "nearly" : "done";
```
jablko, line 387:
```suggestion
        const state: Staleness["state"] = days <= freshDays ? "fresh" : days <= attnDays ? "attention" : days <= nearDays ? "nearly" : "done";
```
makes the `as Staleness` type assertion unnecessary? Too bad it's not inferred.

src/compute-pr-actions.ts
```diff
@@ -394,6 +379,34 @@ export function process(prInfo: PrInfo | BotEnsureRemovedFromProject | BotNoPack
...
+            if (state === "done") {
+                if (doneColumn === "CLOSE") {
+                    context.shouldClose = true;
+                    context.shouldRemoveFromActiveColumns = true;
+                } else {
+                    context.targetColumn = doneColumn;
+                }
+            }
+        };
+        return { kind, days, state, explanation, doTimelineActions } as Staleness;
```
jablko, line 406:
```suggestion
        return { kind, days, state, explanation, doTimelineActions };

File: src/compute-pr-actions.ts (hunk 6 above)
Comment (jablko) [inline src/compute-pr-actions.ts:387]:
```suggestion
        if (days <= freshDays) return undefined;
        const state = days <= attnDays ? "attention" : days <= nearDays ? "nearly" : "done";
```
It looks like `{ kind: "fresh", ... }` and `undefined` are equivalent and the redundant state can be pruned? Just a thought.

elibarzilay: Not needed (see below).

src/compute-pr-actions.ts
elibarzilay: I looked at your `state` suggestion, and the inference is done, just not preserved when put in an object.  So adding an `as const` instead of the above achieves the same purpose too, and IMO it's better than both alternatives.

src/compute-pr-actions.ts
elibarzilay: In theory, you're right -- and this is because a `fresh` state means that nothing is done, so it could have been dropped.

But in practice, there are two reasons I still kept it:

1. Currently, it's useful as a debugging tool: just looking at the `staleness` shows what timeline a PR is headed too, even if it has no implications ATM.  (And yes, there's no direct way currently to view this information, I'm working on a modification of the `daily` script that will show that too.)

2. In the future, the staleness could be used as another bit of information -- it just happens not to be useful right now.  For example,

   * `info.staleness?.kind === "YSYL"` can replace `info.canBeSelfMerged && info.mergeOfferDate`

   * `info.staleness?.kind === "Abandoned"` can replace `info.needsAuthorAction` (because it implies no `canBeSelfMerged`)

   (This could also happen with new staleness timelines too.)

   In any case, ATM, there's only the second one that could be done now.  I hesitate to do so, but I left a comment about it instead.  The reason I think this could be good in the future, is to avoid repeating information in a way that can lead to inconsistencies.  (Also, it's kind of a mental shift from a question of "does it need an action" to "is it on a path for abandonment".)

src/compute-pr-actions.ts
jablko: Ah that makes sense. Nice :+1:

src/compute-pr-actions.ts
jablko: I see. Replacing `needsAuthorAction` and `canBeSelfMerged` w/ `staleness` would make it more explicit that these states aren't indefinite. Seems good!

andrewbranch (approved):

PR #243, opened 2020-11-30
Implement a retry mechanism for PR queries

elibarzilay:
This fixes the problem where an `UNKNOWN` value leads to the bot
toggling the conflicts label.  (See for example
DefinitelyTyped/DefinitelyTyped#48723.)

With this, it should be practical to make switch the daily run to once
every four hours (which I thought that I did a while ago).

### DIFF
--- a/src/pr-info.ts
+++ b/src/pr-info.ts
@@ -163,14 +163,29 @@ function getHeadCommit(pr: GraphqlPullRequest) {
 
 // Just the networking
 export async function queryPRInfo(prNumber: number) {
-    return await client.query<PRQueryResult>({
-        query: GetPRInfo,
-        variables: {
-            pr_number: prNumber
-        },
-        fetchPolicy: "network-only",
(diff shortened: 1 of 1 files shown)


andrewbranch (approved):
Looks good, though is this the only query that needs retries? The retry logic could be extracted to a function so it can wrap any of the queries. But if that’s not needed, this is good.

elibarzilay:
@andrewbranch Yeah, I was worried about this too.  I asked GH about the whole thing, and it sounds like the `mergeable` field is the only thing to worry about (and that's only in this query, which is used only in `pr-info`).  So it looks like it's fine to do this just here.

(I'm guessing that they're running some git operation in the backend to see if it can do a clean merge, and put the result on a branch with some process that cleans up these branches after a (short) while -- that's probably the same thing with the GH "This branch has no conflicts" thing appearing after a few seconds when viewing a PR.)

PR #244, opened 2020-11-30
Run script

elibarzilay:
Two commits:

## Reorganize `package.json` etc

* Reorganize values to make a bit more sense.
* Add and edit some values (description, edit urls to the actual repo).
* Not that it used, but bump the version.
* Update `package-lock.json`.
* Move `@types/jest` into `devDependencies`.

## New `run` script

There's a bunch of stuff in `commands` and `scripts`, but it's a mess to
update all of them.  Instead, this is a single `run` script that can be
used as a drop-in replacement for `daily`, which can do a dry run, show
various info (including the new `ExtendedPrInfo`) in multiple formats,
and can also operate on a selection of PRs.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,17 +1,11164 @@
 {
   "name": "dt-mergebot",
-  "version": "1.0.9",
-  "lockfileVersion": 1,
+  "version": "2.0.0",
+  "lockfileVersion": 2,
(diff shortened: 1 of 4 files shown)


andrewbranch (approved):

orta:
Very cool

PR #254, opened 2020-12-07
Fix unreviewed staleness timeline

elibarzilay:
Two changes:

* The `Unreviewed` timeline needs `!(requireMaintainer && !blessable)`
  to start.

* `noOtherOwners` => `requireMaintainer`.

Also:

* Reorganize the subterms in `blessable` to match `requireMaintainer` so
  the relationship between the two is clearer.

* Change `Comments.StalenessComment` so the index returns a
  function (instead of the other way around), which makes it possible to
  spit out an error line if it uses `otherOwners` and it is empty.

* Add this error to `comments.ts`.  Note that it doesn't throw an
  exception, just prints the error.  This is because if it happens, it's
  not a big problem to mention `«anyone?»`, but it is useful to debug
  such cases (which shouldn't happen, and with this fix they don't).

Fixes the new problem that was raised in #105.

### DIFF
--- a/src/compute-pr-actions.ts
+++ b/src/compute-pr-actions.ts
@@ -149,8 +149,8 @@ function extendPrInfo(info: PrInfo): ExtendedPrInfo {
     const hasTests = info.pkgInfo.some(p => p.files.some(f => f.kind === "test"));
     const newPackages = noNullish(info.pkgInfo.map(p => p.kind === "add" ? p.name : null));
     const hasNewPackages = newPackages.length > 0;
-    const requireMaintainer = editsInfra || editsConfig || hasMultiplePackages || !hasTests || hasNewPackages || tooManyOwners;
-    const blessable = !(hasNewPackages || editsInfra || noOtherOwners);
+    const requireMaintainer = editsInfra || hasNewPackages || noOtherOwners || editsConfig || hasMultiplePackages || !hasTests || tooManyOwners;
+    const blessable = !(editsInfra || hasNewPackages || noOtherOwners);
     const approvedReviews = info.reviews.filter(r => r.type === "approved") as ExtendedPrInfo["approvedReviews"];
     const changereqReviews = info.reviews.filter(r => r.type === "changereq") as ExtendedPrInfo["changereqReviews"];
     const staleReviews = info.reviews.filter(r => r.type === "stale") as ExtendedPrInfo["staleReviews"];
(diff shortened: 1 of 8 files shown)


src/compute-pr-actions.ts
```diff
@@ -183,7 +183,7 @@ function extendPrInfo(info: PrInfo): ExtendedPrInfo {
             "Unmerged", info.mergeOfferDate, 4, 9, 30, "CLOSE");
         if (needsAuthorAction) return mkStaleness(
             "Abandoned", info.lastActivityDate, 6, 22, 30, "CLOSE");
-        if (!approved) return mkStaleness(
+        if (!approved && !(requireMaintainer && !blessable)) return mkStaleness(
```
jablko, line 186: It looks like this change (`noOtherOwners` => `requireMaintainer`) would change some PRs from `approverKind` "other" to `approverKind` "maintainer". What about the following instead?
```suggestion
        if (!approved && approverKind !== "maintainer") return mkStaleness(
```

elibarzilay: That's a good point: made me notice that I can simplify `getApproverKind` a lot with that change.

But I'm not following the purpose of the suggestion: that line has no effect on `approverKind`, so what's the purpose?  (Not objecting to the change since if a maintainer is required, then the `Unreviewed` timeline is just a cosmetic change since it'll be in the maintainer queue regardless, so I very slightly prefer my line just to minimize test (& behavior) changes.)

jablko: I wasn't clear, sorry. I'm suggesting this line instead of adding `noOtherOwners` to `requireMaintainer`, which *does* affect `approverKind` (specifically cases that were previously `approverKind` "other" and are now `approverKind` "maintainer").

elibarzilay: Ah, well -- the change in this case is *intended*...  Just in case, I looked inot it to verify that it works fine: the changed `approverKind` happens in two tests: 44282 and 49548 -- in both of these, there is no other owner, so it was bogus to have it get into a waiting-for-review in the first test, and get into the "unreviewed" stale state in the second.

Another way to see the problem is that with your suggestion, the `console.error` that I added to warn when there are no other owners *does* happen (for the unreviewed case, since it's trying to ping the other owners when there are none).

(It's true that there might be a point in waiting for a few days in a waiting-for-review state, in case someone steps in and provides a review, I'll ask about it.)

orta: I think generally making it auto switch is probably fine. If there was someone else who would give any sort of useful review in a reasonable timeframe I'd imagine they'd get added pretty soon?

Otherwise seems good to me 👍🏻

orta (approved):

src/compute-pr-actions.ts
jablko: When `approverKind` would be "owner" except that `noOtherOwners`, already that makes the `approverKind` "maintainer", which clearly it should be. +1 for fixing the unreviewed timeline in that case.

Changing `approverKind` "other" to `approverKind` "maintainer" feels like a different change. Not that I'm opposed, but if we do make that change, it should be uniform? It feels weird to me to make `approverKind` "maintainer" if there's one owner and "other" if there are multiple. Or "maintainer" if the owner made the change and "other" if someone else did?

PR #255, opened 2020-12-07
Don't suggest dismissing reviewes

### LINKED ISSUE #253: The “dismiss review” feature is only available to organisation members and collaborators
The bot shouldn’t suggest using features that aren’t available to the creator of the PR: <https://github.com/DefinitelyTyped/dt-mergebot/blob/57cb3a6263820a21efb5f7fd5bb15a815a853dc9/src/comments.ts#L22-L25>

---

**Refs:** <https://docs.github.com/github/collaborating-with-issues-and-pull-requests/dismissing-a-pull-request-review>

elibarzilay:
Fixes #253.

### DIFF
--- a/src/_tests/fixtures/43695-duplicate-comment/_response.json
+++ b/src/_tests/fixtures/43695-duplicate-comment/_response.json
@@ -871,7 +871,7 @@
                 "login": "typescript-bot",
                 "__typename": "User"
               },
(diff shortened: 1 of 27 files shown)


orta (commented):
Yeah, this makes sense. Only those with write access get that button 👍🏻

PR #261, opened 2020-12-08
Make a way to mark comment tags that should be removed if not in actions

elibarzilay:
Use it instead of `actions.isReadyForAutoMerge`.

Also, a random fix in `fetchFile`: `GetFileContentResult` can be null.


orta (approved):
Cool, makes sense

PR #262, opened 2020-12-08
Suggest testing in a comment

elibarzilay:
## Add `hasDefinitions` and `isUntested`

* `hasDefinitions` is used to make `isUntested` indicate having
  definition changes but no tests -- better than `!hasTests` which would
  be true for other kinds of changes.

* Use it for the "Untested Change" label, and make the welcome comment
  not say "and updated the tests" if, for example, there were only test
  changes.

## Make the test suggestion text be a comment

Instead of having it in the welcome message where it is much less
visible.  Add an emoji to make it stand out even more.

Note that the comment uses a fixed tag, and it is removed if tests are
added.  This is to minimize the extra nagging.


src/compute-pr-actions.ts
```diff
@@ -479,7 +482,8 @@ function createWelcomeComment(info: ExtendedPrInfo) {
     } else if (info.popularityLevel === "Critical" && !info.maintainerBlessed) {
         display(`Because this is a widely-used package, ${requiredApprover} will need to review it before it can be merged.`);
     } else if (!info.requireMaintainer) {
-        display("Because you edited one package and updated the tests (👏), I can help you merge this PR once someone else signs off on it.");
+        const and = info.hasDefinitions && info.hasTests ? "and updated the tests (👏)" : "(and no type definition changes)";
```
rbuckton, line 485:
```suggestion
        const and = info.hasDefinitions && info.hasTests ? "and updated the tests (👏)" : "and there were no type definition changes";

```

rbuckton (changes requested):

PR #268, opened 2020-12-11
Drop `No Other Owners` on infra PRs

elibarzilay:
* New `hasEditedPackages`: which is true when there *are* packages that
  are not new.  (I.e., not counting infra changes.)

* Use that as a condition for `No Other Owners`, so it doesn't turn on
  for infra PRs.

* Do the same for `Config Edit`, though there are no tests affected.

Supersedes #267.

### DIFF
--- a/src/compute-pr-actions.ts
+++ b/src/compute-pr-actions.ts
@@ -128,6 +128,7 @@ export interface ExtendedPrInfo extends PrInfo {
     readonly isUntested: boolean;
     readonly newPackages: readonly string[];
     readonly hasNewPackages: boolean;
+    readonly hasEditedPackages: boolean;
     readonly needsAuthorAction: boolean;
     readonly reviewColumn: ColumnName;
     readonly isAuthor: (user: string) => boolean; // specialized version of sameUser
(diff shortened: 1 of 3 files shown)


src/compute-pr-actions.ts
```diff
@@ -284,10 +286,10 @@ export function process(prInfo: BotNotFail,
     label("Edits Infrastructure", info.editsInfra);
     label("Edits multiple packages", info.hasMultiplePackages);
     label("Author is Owner", info.authorIsOwner);
-    label("No Other Owners", !info.hasNewPackages && info.noOtherOwners);
+    label("No Other Owners", info.hasEditedPackages && info.noOtherOwners);
     label("Too Many Owners", info.tooManyOwners);
     label("Self Merge", info.canBeSelfMerged);
-    label("Config Edit", !info.hasNewPackages && info.editsConfig);
+    label("Config Edit", info.hasEditedPackages && info.editsConfig);
```
jablko, line 292: Beyond the scope of the original change but maybe?
```suggestion
    label("Config Edit", info.editsConfig);

jablko (approved):
I'm in favor, thanks!

I didn't like the non-equivalence between `noOtherOwners` and the homonymous label, but as you pointed out, the non-equivalence between `noOtherOwners` and `otherOwners.length === 0` is worse. Thanks for this change!

Beyond the scope of the original change but speaking of "Config Edit", I was surprised that label is excluded from new definition PRs, because it provides a bit of information: whether the new definition adheres to the required form. What do think about dropping the `hasEditedPackages` condition there?

elibarzilay:
> I didn't like the non-equivalence between `noOtherOwners` and the homonymous label, but as you pointed out, the non-equivalence between `noOtherOwners` and `otherOwners.length === 0` is worse. Thanks for this change!

Yeah -- I generally view the info fields as things that should be straightforward, and the labels as the displayed information where it might make sense to have a property on one side that is different on the other side..


> Beyond the scope of the original change but speaking of "Config Edit", I was surprised that label is excluded from new definition PRs, because it provides a bit of information: whether the new definition adheres to the required form. What do think about dropping the `hasEditedPackages` condition there?

That's a bigger change.  "Edits config" started with just that meaning, and then it changed a bit, and this would change it even more.  For example "New Package + Edits Config" seem weird enough to warrant changing it to something like "Suspicious Config Edit".  It might be a good idea, but I'd like that to be a separate thing.  Maybe filed as an issue?

PR #274, opened 2020-12-12
Add link for `tslint.json` contents

elibarzilay:
Supersedes #223.

Co-authored-by: Jack Bates <jack@nottheoilrig.com>

### DIFF
--- a/src/pr-info.ts
+++ b/src/pr-info.ts
@@ -404,7 +404,8 @@ configSuspicious["package.json"] = makeJsonCheckerFromCore(
 );
 configSuspicious["tslint.json"] = makeJsonCheckerFromCore(
     { extends: "dtslint/dt.json" },
-    []
+    [],
+    urls.linterJson
 );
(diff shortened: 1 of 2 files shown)


jablko:
What do you think about keeping the most important bit of the explanation as in the latest revision to #223, like https://github.com/DefinitelyTyped/DefinitelyTyped/pull/49575/commits/cdd66f483bb0103f6e6d1eb7b07192ff95718f25?

elibarzilay:
> I've updated #233: What do you think about keeping the most important bit of the explanation, like [DefinitelyTyped/DefinitelyTyped@cdd66f4](https://github.com/DefinitelyTyped/DefinitelyTyped/commit/cdd66f483bb0103f6e6d1eb7b07192ff95718f25)?

I have a strong preference to keeping that code from blowing up, and be as consistent as possible.  Adding the additional argument makes it overengineered (for a tiny benefit), and the information is already there.

jablko:
That makes sense. I'll close #223

PR #284, opened 2020-12-19
Fix `README.md` links

elibarzilay:
Looks like `#foo` anchors work when viewing the DT page, for for links
from other pages, the anchors usually fail.  Looking at the actual
rendered HTML, the ids all start with `user-content-`.  So add that, and
do so via a function if they change that in the future.

Also, use the two links for the section on tesing new/existing packages.


orta:
👍🏻

src/urls.ts
```diff
@@ -1,9 +1,12 @@
 const baseURL = "https://github.com/DefinitelyTyped/DefinitelyTyped";
+const readmeLink = (header: string) =>
+    `${baseURL}#user-content-${header.replace(/[^\w -]/g, "").replace(/ /g, "-").toLowerCase()}`;
```
jablko, line 3: Alternatively we could link to e.g. https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#tsconfigjson where the section links that GitHub exposes work without the `user-content-` workaround?

PR #285, opened 2020-12-19
Always show the list of files to check

elibarzilay:
Motivated by the recent change of showing the `Check Config` label also
for new packages, where in such cases the usual place where the
suspicious configs are shown is not reached since there's an earlier
case that shows a different comment for new packages.  Instead of
showing the list of files there too, this just moves the whole thing
into the packages display section (which shows the links and pre-package
approvals).


elibarzilay:
For example, see below for how DefinitelyTyped/DefinitelyTyped#50159 is shown with this.  Note also that this makes it a bit more visible, and I'm guessing that we'll need to improve the code as a result -- in this case, the complaint is because there's a `"DOM"` added to `compilerOptions.lib`.

image

elibarzilay:
@jablko: I didn't check but this is likely to conflict with the big PR for adding suggestions-as-reviews, but I think that it's a good idea to use this as a first step in adding it gradually:
* Start with this, which shows the list in a consistent place so people will see it more.
* Would be nice to then add a short one-line explanation for each bit that made the change suspicious (in the above case, the `DOM` addition).
* This would make it even more useful (for both contributors and maintainers), and will probably lead to complaints about things like this `dom` thing where the bot should do a better job.  (I think that this is a very good thing to do before the last step :arrow_right:)
* Add the suggestion reviews.  (:arrow_right: because once it's a review, it should be very robust since many people will probably just blindly accept whatever the bot suggest, and will be annoyed if it's incorrect.)

orta (approved):
I agree with the ideas 👍🏻 - implementation looks good too

src/_tests/fixtures/45946/mutations.json
```diff
@@ -35,7 +35,7 @@
...
...thor is owner)\n  - Config files to check:\n    - [`asynciterator/tsconfig.json`](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45946/files/a5eb1f19e7a31387e7d7c79b3acc483b0e60b32e#diff-be20504719e5f1036e683d4ddaab8bf2b1c609ea12aa30786497768f575095ad): couldn't fetch contents\n    - [`asynciterator/tslint.json`](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45946/files/a5eb1f19e7a31387e7d7c79b3acc483b0e60b32e#diff-e3a26c4ba64c5ed587d2b797151728488e274f4be74c1558722a8a32cf7a31c9): couldn't fetch contents\n\n## Code Reviews\n\nThere aren't any other owners of this package, so a DT maintainer will review it.\n\n## Status\n\n * ✅ No merge conflicts\n * ✅ Continuous integration tests have passed\n * ❌ A DT maintainer needs to approve changes which affect DT infrastructure ([`notNeededPackages.json`](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45946/files/a5eb1f19e7a31387e7d7c79b3acc483b0e60b32e#diff-a275c2eae7b8f788a52327e76809026ee1bc5ea614393806c8f3cbab07202621))\n\nOnce every item on this list is checked, I'll ask you for permission to merge and publish the changes.\n\n----------------------\n... diagnostics scrubbed ...\n<!--typescript_bot_welcome-->"
```
jablko, line 38: Can we land #222 to avoid showing this for removed files? I think #222 is ready to merge?

jablko:
Can we put the explanation next to the line in question? I'll prepare a PR based on #265, minus the actual suggestions.

src/_tests/fixtures/45946/mutations.json
elibarzilay: (I'll try to get to it soon...)

PR #291, opened 2020-12-23
Remove a bunch of outdated scripts

elibarzilay:
Many of these are superseded by `run.ts` (showing info, actions,
mutations; dry runs; operate on a specified PR; cleanup columns), and
are no longer needed.  Also shuffled the list of scripts in
`package.json` a bit.

(There are some more things that are not needed, but this is good for
now...)


src/commands/clean-columns.ts
```diff
@@ -1,60 +0,0 @@
...
-
-const headers = {
-    "User-Agent": "RyanCavanaugh ColumnCleaner",
-    "Authorization": `token ${process.env['AUTH_TOKEN']}`,
-    "Accept": "application/vnd.github.inertia-preview+json"
-}
-
-
-let cleanCount = 0;
-fs.readFile(path.join(__dirname, '../clean-columns-query.graphql'), { encoding: 'utf-8' }, (err, query) => {
```
jablko, line 39: Nice! I think this is the only reference to https://github.com/DefinitelyTyped/dt-mergebot/blob/master/clean-columns-query.graphql so we can add that file to this PR?

elibarzilay: Bah, you're right -- I wanted to check that one but forgot...

PR #307, opened 2021-01-03
Test updates

elibarzilay:
* Mainly switch from `github:Microsoft/definitelytyped-header-parser` to
  `@definitelytyped/header-parser` to avoid the annoying
  `flatmap-stream` not found error when installing.

* Regenerate `package-lock.json`

* Some more updates
  (As long as I'm going over this...)


elibarzilay:
(Merging now, since I lost my old `node_modules` and cannot reinstall it due to the old parser header dependency.)

PR #308, opened 2021-01-04
No need to keep `headCommitAbbrOid` in the (extended) info

elibarzilay:
All uses are within `process`, so define it there.


jablko (approved):
Nice!

PR #315, opened 2021-01-06
Fix CI status checking

andrewbranch:
The new merge codeowners GH Action broke the existing logic.


orta:
Ah, thanks - this makes sense!

PR #317, opened 2021-01-07
Convert `PR-Trigger/index.js` to TS

elibarzilay:
* Removed `sideEffects` and instead just try it for the single entry
  that was there.

* Refactored things a bit around `acceptedEventsToActions` and `allowListedActions`

* Dropped `if (state.type === "noop")` (similar to #294)

* Some touchups in `src/side-effects/merge-codeowner-prs`


PR-Trigger/index.ts
```diff
@@ -37,34 +34,31 @@ const httpTrigger = async function (context, _req) {
 
     // Allow the bot to run side-effects which are not the 'core' function
     // of the review cycle, but are related to keeping DT running smoothly
-    const sideEffects = {
-        "check_suite": mergeCodeOwnersOnGreen
+    if (event === "check_suite") {
+        await mergeCodeOwnersOnGreen(req.body as EventPayloads.WebhookPayloadCheckSuite);
```
orta, line 38: Good call on some of these simplifications - now we know what this codebase is, we don't really need it

PR-Trigger/index.ts
```diff
@@ -37,34 +34,31 @@ const httpTrigger = async function (context, _req) {
...
-        "check_suite": mergeCodeOwnersOnGreen
+    if (event === "check_suite") {
+        await mergeCodeOwnersOnGreen(req.body as EventPayloads.WebhookPayloadCheckSuite);
     }
 
-    if (sideEffects[event]) sideEffects[event](req.body);
-
     // https://developer.github.com/webhooks/
-    const acceptedEventsToActions = {
+    const acceptedEventsToActions: Record<string, string[] | undefined> = {
```
orta, line 42: Heh, you can always turn on http://typescriptlang.org/tsconfig#noUncheckedIndexedAccess

orta (approved):
Looks good to me

PR-Trigger/index.ts
elibarzilay: 👍

PR #323, opened 2021-01-13
fix(tsconfig.json): Ignore changes to the `lib` field

ExE-Boss:
Fixes <https://github.com/DefinitelyTyped/dt-mergebot/issues/318>
Supersedes and closes <https://github.com/DefinitelyTyped/dt-mergebot/pull/319>


elibarzilay:
Asked around, and there's support for allowing only `es6` and `dom`.

PR #325, opened 2021-01-13
Move `PR-Trigger/index.ts` to `src/pr-trigger.ts`

elibarzilay:
Also:
* Switch back to plain `tsconfig.json`
* Hack a quick script to show the tree instead of apt-installing `tree`
* Remove `node_modules` and reinstall only production packages


orta:
Yep, this looks good to go for me - I'll monitor the deploy

PR #326, opened 2021-01-13
Cleanup

elibarzilay:
* Convert all CRLF files to LF (since it became messy with some files on
  each side)
* Remove displaying `pr-trigger.js` again

(I realize that this can mess PRs -- I can/will do the rebasing when needed.)


jablko (approved):
:100:

> (I realize that this can mess PRs -- I can/will do the rebasing when needed.)

In the past I've found `--strategy-option ignore-cr-at-eol` helpful in such cases, as in
```Shell
git pull --strategy-option ignore-cr-at-eol upstream master
```

elibarzilay:
Good hint!  -- And since you're almost the only sufferer from this, I'll merge it...

PR #334, opened 2021-01-16
Better date code

elibarzilay:
Switch to dayjs, send the expiration string to `StalenessComment`, use
ordinal days.


src/compute-pr-actions.ts
```diff
@@ -375,8 +378,7 @@ function makeStaleness(now: string, author: string, otherOwners: string[]) { //
         const state = days <= freshDays ? "fresh" : days <= attnDays ? "attention" : days <= nearDays ? "nearly" : "done";
         const kindAndState = `${kind}:${state}`;
         const explanation = Comments.StalenessExplanations[kindAndState];
-        const expires = new Date(now);
-        expires.setDate(expires.getDate() + nearDays);
```
jablko, line 379: That works too, good idea to use the library :+1:

jablko (approved):

PR #335, opened 2021-01-16
More cleanups

elibarzilay:
* Indentation, semicolons, quotes, etc.

* Eslint rules to check these (ignoring some things that I couldn't
  figure out a way to sensible-ize).

* Drop the `CIResult` enum and use the strings directly instead.


jablko (approved):
These all look like improvements to me :+1:

PR #380, opened 2021-03-12
Used multi-state emojis for CI result comments

### LINKED ISSUE #365: Consider ⏳ emoji instead of ❌  for pending
Hi! I love the live comment feature on DT pull requests. It's very useful.

Today was the first time I looked at it immediately after sending a PR and was alarmed for a brief moment at the ❌ s:
 
Screenshot of typescript-bot informatively commenting on https://github.com/DefinitelyTyped/DefinitelyTyped/pull/51228#issuecomment-778855433

Could we switch those to, say, ⏳  to indicate the task is _pending_, not _failed_? I'd be happy to make that change if you're up for it!

JoshuaKGoldberg:
Fixes #365

These existing test cases are 👌...


src/compute-pr-actions.ts
```diff
@@ -487,12 +488,12 @@ function createWelcomeComment(info: ExtendedPrInfo, post: (c: Comments.Comment)
     display(``,
             `## Status`,
             ``,
-            ` * ${emoji(!info.hasMergeConflict)} No merge conflicts`);
+            ` * ${emoji.failure(!info.hasMergeConflict)} No merge conflicts`);
```
elibarzilay, line 491: This is kind of confusing to read, would be better if the argument was flipped, so `emoji.failure(x)` means failure if `x` is true.

src/compute-pr-actions.ts
```diff
@@ -487,12 +488,12 @@ function createWelcomeComment(info: ExtendedPrInfo, post: (c: Comments.Comment)
...
 
-    const expectedResults = info.ciResult === "unknown" ? "finished" : "passed";
-    display(` * ${emoji(info.ciResult === "pass")} Continuous integration tests have ${expectedResults}`);
+    const result = emoji.result(info.ciResult);
+    display(` * ${result.emoji} Continuous integration tests ${result.text}`);
```
elibarzilay, line 494: Maybe do this since `result` is a very generic name?

```suggestion
    { const result = emoji.result(info.ciResult);
      display(` * ${result.emoji} Continuous integration tests ${result.text}`); }
```

src/compute-pr-actions.ts
```diff
@@ -487,12 +488,12 @@ function createWelcomeComment(info: ExtendedPrInfo, post: (c: Comments.Comment)
...
 
-    const approved = emoji(info.approved);
+    const approved = emoji.pending(info.approved);
```
elibarzilay, line 496: This is confusing like `emoji.failure` -- I think that `emoji.pending(!info.approved)` would be clearer.

src/emoji.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export function result(result: CIResult) {
+    switch (result) {
+        case "fail":
+            return { emoji: "❌", text: "have failed" };
+        case "pass":
+            return { emoji: "✅", text: "have passed" };
+        case "unknown":
+            return { emoji: "🕐", text: "are still running" };
+        default:
```
elibarzilay, line 19: Should be `case "missing"` so there's a type error if the type gets extended in the future.

src/emoji.ts
```diff
@@ -0,0 +1,22 @@
...
+    return passed ? "✅" : "❌";
+}
+
+export function pending(passed: boolean) {
+    return passed ? "✅" : "🕐";
+}
+
+export function result(result: CIResult) {
+    switch (result) {
+        case "fail":
```
elibarzilay, line 13: General pickiness: no need for newlines before `return`s?

src/emoji.ts
```diff
@@ -0,0 +1,22 @@
+import { CIResult } from "./pr-info";
+
+export function failure(passed: boolean) {
```
elibarzilay, line 3: General pickiness: I prefer `const f = x => ...` for small functions.

elibarzilay (changes requested):

src/compute-pr-actions.ts
JoshuaKGoldberg: I ended up switching to `.passed` to match up with the other functions. But yes 😄 thanks!

src/compute-pr-actions.ts
JoshuaKGoldberg: This might be the first time in my career I've intentionally used a standalone block. 😄

elibarzilay (changes requested):
Ugh, I missed this earlier: it's probably better to use `:xxx:` syntax instead of unicode...?

(Just ignore if you agree, I'll change it when merging.)

elibarzilay (approved):

PR #386, opened 2021-04-26
Moves PRs which need manually accepting for CI to the DT maintainer action queue

orta:
New contributors to DT need to have their CI privileges set - this PR changes the DT mergebot to move those PRs directly into the maintainer's TODO list.

## Major additions:

* Switch from a GQL mutation to a REST api call for rerunning builds.
  See the source for the HACK comment re doing a REST call.

* Collect the IDs of *all* such tests, so they're all triggered, rather
  than just the first found.

* New `The CI is blocked` label if test runs need to be approved and
  this cannot be done automatically due to infra edits.

* When there are no infra edits, approve the build (by re-running it),
  and treat it as a test-in-progress.  In this case, mostly treat the CI
  state as `unknown` (still running) for the welcome message etc since
  it'll be started now.

* Add an `info.tooManyFiles` field that tracks when we didn't see all of
  the files, and:
  - `info.possiblyEditsInfra` is true if such edits exist *or* there are
    too many files so we can't be sure.  New `Possibly Edits
    Infrastructure` label reflects this.
  - add a prominent note to the welcome message when there are too many
    files, to alert DT maintainers.


orta:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/52586 would remove the other GH actions CI step which is what is allowing DT mergebot to think it's green right now

andrewbranch:
To automatically approve, you can just rerequest each check suite whose `conclusion` is `ACTION_REQUIRED`: https://docs.github.com/en/graphql/reference/mutations#rerequestchecksuite

I would recommend doing this on the condition that no files outside the `types/` directory are added/removed/modified.

You can also do the same thing with the [REST API](https://docs.github.com/en/rest/reference/checks#rerequest-a-check-suite) if for some reason that’s easier, but you’ll need to use the `databaseId` property instead of the `id` property of the check suite as returned from the GraphQL API.

There’s also another way to do this at the [repo/actions/runs level with the REST API](https://docs.github.com/en/rest/reference/actions#re-run-a-workflow), but that’s not available via GraphQL mutation, and it’s also very difficult to figure out the check run id from the check suite (I don’t think it can even be retrieved with GraphQL at all), so I don’t recommend that.

src/_tests/fixtures/52579-edits-infra/_files.json
```diff
@@ -0,0 +1,5 @@
+{
+  "5337ea17b2407707bead503cd20c9ca29438f5df:types/vertx__eventbus-bridge-client.js/tsconfig.json": "{\n    \"compilerOptions\": {\n        \"module\": \"commonjs\",\n        \"lib\": [\n            \"es6\"\n        ],\n        \"noImplicitAny\": true,\n        \"noImplicitThis\": true,\n        \"strictFunctionTypes\": true,\n        \"strictNullChecks\": true,\n        \"baseUrl\": \"../\",\n        \"typeRoots\": [\n            \"../\"\n        ],\n        \"types\": [],\n        \"noEmit\": true,\n        \"forceConsistentCasingInFileNames\": true\n    },\n    \"files\": [\n        \"index.d.ts\",\n        \"vertx__eventbus-bridge-client.js-tests.ts\"\n    ]\n}\n",
+  "5337ea17b2407707bead503cd20c9ca29438f5df:types/vertx__eventbus-bridge-client.js/tslint.json": "{ \"extends\": \"dtslint/dt.json\" }\n",
+  "5337ea17b2407707bead503cd20c9ca29438f5df:package.json": "{}"
+}
```
orta, line 5: This first fixture is my modified one - it changes the package.json - which will show up as "edits infra" later

src/_tests/fixtures/52579-edits-infra/mutations.json
```diff
@@ -0,0 +1,22 @@
...
+        "body": "@Phong6698 Thank you for submitting this PR! I see this is your first time submitting to DefinitelyTyped 👋 — I'm the local bot who will help you through the process of getting things through.\n\n***This is a live comment which I will keep updated.***\n\nThis PR touches some part of DefinitelyTyped infrastructure, so a DT maintainer will need to review it. This is rare — did you mean to do this?\n\n## 1 package in this PR\n\n* `vertx__eventbus-bridge-client.js` (*new!*) [on npm](https://www.npmjs.com/package/@vertx/eventbus-bridge-client.js), [on unpkg](https://unpkg.com/browse/@vertx/eventbus-bridge-client.js@latest/)\n\n## Code Reviews\n\nThis PR adds a new definition, so it needs to be reviewed by a DT maintainer before it can be merged.\n\n## Status\n\n * ✅ No merge conflicts\n * 🕐 Continuous integration tests waiting on a maintainer accepting CI to run\n * 🕐 Only a DT maintainer can approve changes when there are new packages added\n\nOnce every item on this list is checked, I'll ask you for permission to merge and publish the changes.\n\n----------------------\n... diagnostics scrubbed ...\n<!--typescript_bot_welcome-->"
+      }
+    }
+  }
```
orta, line 21: The key here is that there _isn't_ a mutation in this list to trigger the CI, and there's this message:

>  tests waiting on a maintainer accepting CI to run

src/_tests/fixtures/52579/_files.json
```diff
@@ -0,0 +1,5 @@
...
+}
```
orta, line 5: This one is the 'real' fixture, only changes types

src/_tests/fixtures/52579/mutations.json
```diff
@@ -0,0 +1,20 @@
...
+    }
+  },
+  {
+    "mutation": "mutation ($input: RerequestCheckSuiteInput!) {\n  rerequestCheckSuite(input: $input) {\n    __typename\n  }\n}\n",
+    "variables": {
+      "input": {
+        "checkSuiteId": "123456",
+        "repositoryId": "MDEwOlJlcG9zaXRvcnk2MDkzMzE2"
+      }
+    }
```
orta, line 18: Here's the mutation @andrewbranch  mentioned above

src/commands/create-fixture.ts
```diff
@@ -53,7 +53,7 @@ export default async function main(directory: string, overwriteInfo: boolean) {
     writeJsonSync(resultFixturePath, actions);
 
     const mutationsFixturePath = join(fixturePath, "mutations.json");
-    const mutations = await executePrActions(actions, prInfo, /*dry*/ true);
+    const mutations = await executePrActions(response.data.repository!.id, actions, prInfo, /*dry*/ true);
     writeJsonSync(mutationsFixturePath, mutations);
```
orta, line 57: The API for re-running requires an ID to the repo, and that's not in the prInfo. Given it's only used here, I figure it made more sense to add it as an arg

andrewbranch (approved):

orta (commented):
Yeah, these changes looks good to me 👍🏻

Ry0taK (approved):
Looks good to me :smile:

src/pr-info.ts
```diff
@@ -472,12 +484,20 @@ function getReviews(prInfo: PR_repository_pullRequest) {
     return reviews;
 }
 
-function getCIResult(checkSuites: PR_repository_pullRequest_commits_nodes_commit_checkSuites | null): { ciResult: CIResult, ciUrl?: string } {
+function getCIResult(checkSuites: PR_repository_pullRequest_commits_nodes_commit_checkSuites | null): { ciResult: CIResult, ciUrl?: string, reRunCheckSuiteIDs?: number[] } {
     const ghActionsChecks = checkSuites?.nodes?.filter(check => check?.app?.name.includes("GitHub Actions"));
+
+    // Freakin' crypto miners ruined GitHub Actions, and now we need to manually confirm new folks can run CI
```
johnnyreilly, line 490: Made me laugh 😂

orta:
Confirming that this worked locally, triggering https://github.com/DefinitelyTyped/DefinitelyTyped/pull/52627 

I'll merge and nurse the PR

src/pr-info.ts
orta: I had to really redact myself, crypto spammers also attacked the typescript discord this week

orta:
Cool, I watched https://github.com/DefinitelyTyped/DefinitelyTyped/pull/52626 get updated by the bot

PR #391, opened 2021-05-06
Fetch *all* files

elibarzilay:
This reduces maintainer load since there are less suspicious PRs, and
make it easier to see if the PR is well behaved.

Move the whole retry + fetch rest thing into `src/queries/pr-query`.

Leave the `tooManyFiles` bits in for now, but if this is stable enough
they should eventually be removed.


orta (approved):
Cool, this looks structurally right to me 👍🏻

PR #396, opened 2021-05-10
Update packages

elibarzilay:
To avoid a new warning, and also update jest to get a more recent
typescript.


orta:
worth running npm deduce too  https://docs.npmjs.com/cli/v6/commands/npm-dedupe

PR #408, opened 2021-05-19
Use the JSON from the main comment to skip doing work

orta:
Re: the API for the playground plugin


orta:
I'm gonna self merge - if it's wrong it'll only break me

PR #409, opened 2021-05-19
Adds a link to the DT Review Playground plugin

orta:
Adds an extra comment with a link to review on the main comment. Blocked on #408 and then deploying the plugin to npm.


src/api.ts
```diff
@@ -1,40 +1,39 @@
...
 
-    const info = await getPRInfo(prNumber)
+    const info = await getPRInfo(prNumber);
     const prInfo = info.data.repository?.pullRequest;
 
-    if (!prInfo)  return notFound("No PR metadata")
-    if (!prInfo.comments)  return notFound("No PR comments")
+    if (!prInfo)  return notFound("No PR metadata");
+
+    const welcomeComment = prInfo.comments.nodes!.find(c =>  c && c.author?.login === "typescript-bot" && c.body.includes("<!--typescript_bot_welcome-->"));
```
elibarzilay, line 30: BUG!  Double spaces in `=>  c`!

But more seriously, I think that it's a bit better to use `endsWith` so it does the same as the comment parsing thing -- or even better, just use *that* (the stuff in `util/comment.ts`).

src/api.ts
```diff
@@ -1,40 +1,39 @@
...
+    if (!welcomeComment || !welcomeComment.body || !welcomeComment.body.includes("```json")) return notFound("PR comment with JSON not found");
 
-    const welcomeComment = prInfo.comments.nodes!.find(c =>  c && c.author?.login === "typescript-bot" && c.body.includes("<!--typescript_bot_welcome-->"))
-    if (!welcomeComment || !welcomeComment.body || !welcomeComment.body.includes("```json")) return notFound("PR comment with JSON not found")
-    
     // Extract the JSON from the comment
-    const jsonText = welcomeComment.body.split("```json")[1]!.split("```")[0]!
+    const jsonText = welcomeComment.body.split("```json")[1]!.split("```")[0]!;
```
elibarzilay, line 34: I have a slight preference to the regexp I used in teams:

    .replace(/^[^]*```json\n([^]*)\n```[^]*$/, "$1")

since it's more straightforward than the `.split`s, and a little stronger preference for that since it will always return the last ```` ```json...``` ```` block (in case more blocks are added later).

elibarzilay (approved):

PR #410, opened 2021-06-03
Notes that we don't look at red builds

orta:
Maybe it's not the most effective copy, but it does get the point across

### DIFF
--- a/src/_tests/fixtures/43160/mutations.json
+++ b/src/_tests/fixtures/43160/mutations.json
@@ -42,7 +42,7 @@
     "variables": {
       "input": {
         "subjectId": "MDExOlB1bGxSZXF1ZXN0Mzg5MDYwOTQ4",
-        "body": "@rikkertkoppes The CI build failed! Please [review the logs for more information](https://github.com/DefinitelyTyped/DefinitelyTyped/commit/22c73c88cc9c09efd4c2998ec360607dd4c36c2e/checks?check_suite_id=731664306).\r\n\r\nOnce you've pushed the fixes, the build will automatically re-run. Thanks!\n<!--typescript_bot_gh-actions-complaint-6d5d2a8-->"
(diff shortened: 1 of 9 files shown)


src/_tests/fixtures/43160/mutations.json
```diff
@@ -42,7 +42,7 @@
     "variables": {
       "input": {
         "subjectId": "MDExOlB1bGxSZXF1ZXN0Mzg5MDYwOTQ4",
-        "body": "@rikkertkoppes The CI build failed! Please [review the logs for more information](https://github.com/DefinitelyTyped/DefinitelyTyped/commit/22c73c88cc9c09efd4c2998ec360607dd4c36c2e/checks?check_suite_id=731664306).\r\n\r\nOnce you've pushed the fixes, the build will automatically re-run. Thanks!\n<!--typescript_bot_gh-actions-complaint-6d5d2a8-->"
+        "body": "@rikkertkoppes The CI build failed! Please [review the logs for more information](https://github.com/DefinitelyTyped/DefinitelyTyped/commit/22c73c88cc9c09efd4c2998ec360607dd4c36c2e/checks?check_suite_id=731664306).\r\n\r\nOnce you've pushed the fixes, the build will automatically re-run. **Note: builds which are failing do not end up on the DT maintainers list of PRs to review**. Thanks!\n<!--typescript_bot_gh-actions-complaint-6d5d2a8-->"
```
RyanCavanaugh, line 45: maintainers'

src/comments.ts
```diff
@@ -18,7 +18,7 @@ export const HadError = (user: string | undefined, error: string) => ({
 
 export const CIFailed = (abbrOid: string, user: string, ciUrl: string) => ({
     tag: `gh-actions-complaint-${abbrOid}`,
-    status: `@${user} The CI build failed! Please [review the logs for more information](${ciUrl}).\r\n\r\nOnce you've pushed the fixes, the build will automatically re-run. Thanks!`
+    status: `@${user} The CI build failed! Please [review the logs for more information](${ciUrl}).\r\n\r\nOnce you've pushed the fixes, the build will automatically re-run. **Note: builds which are failing do not end up on list of PRs for DT maintainers to review.** Thanks!`
```
elibarzilay, line 21: Shouldn't there be an empty line before this so it's on its own line?

(Assuming that you're out for the day I'll just do that...)

PR #411, opened 2021-06-04
Add a utility to wrap long lines of text

elibarzilay:


orta (approved):
seems reasonable - there's https://github.com/cspotcode/outdent if you want to have a dep to do this instead too

elibarzilay:
Yeah, I know about things like {de,out,un}dent and related variants...  What I added was specifically a cheap & small thing and a bit less magical...

(I have a [long history](https://www.barzilay.org/misc/scribble-reader.pdf) in this area, see for example [these things](https://docs.racket-lang.org/scribble-pp/text.html).)

PR #412, opened 2021-06-04
Adds a general shape for discussions support

orta:
Adds a new function route for discussions on the DT repo.

Main features:
- When you ask for help on a dt module then it requires you have `[module-name]` in the title somewhere, then it pings everyone and sets a label on the discussion for you. If you don't it asks you for info. If you don't it asks for it.

Some things that this means:

 - We will have categorized discussions which haven't been hooked up via a label
 - People can keep see overall what's going on with a dep via the labels 
 - It doesn't rely on folks doing the work to find the right people to ping, which doesn't seem to happen too often

Future: When you post a new request for a dt module, it pings the discord


orta:
OK, I think this is pretty ready to go

src/discussions-trigger.ts
```diff
@@ -0,0 +1,175 @@
...
+const couldNotFindMessage = `Hi, we could not find a reference to the types you are talking about in this discussion. 
+Please edit the title to include the name on npm inside square brackets. 
+
+E.g. 
+ - \`"[@typescript/vfs] Does not x, y"\`
+ - \`"Missing x inside [node]"\`
+ - \`"[express] Broken support for template types"\`
+`;
+
+const gotAReferenceMessage = (module: string, owners: string[]) => `Thanks for the discussion about "${module}", some useful links for everyone: [npm](https://www.npmjs.com/package/${module}) / etc
```
elibarzilay, line 71: (And this would reuse the bits from `compute-pr-actions.ts`, I'm guessing, possibly the popularity labels too.)

src/discussions-trigger.ts
```diff
@@ -0,0 +1,175 @@
...
+
+
+async function pingAuthorsAndSetUpDiscussion(discussion: Discussion) {
+    const aboutNPMRef = extractNPMReference(discussion);
+    if (!aboutNPMRef) {
+        // https://gist.github.com/smashwilson/311e1487ddb40a455fc54d294cc63ad4#adddiscussioncomment
+        await updateOrCreateMainComment(discussion, couldNotFindMessage);
+        return;
+    } else {
+        const message = gotAReferenceMessage(aboutNPMRef, []);
```
elibarzilay, line 84: So this is still pretty rough?  If it's merged as-is, then there's the missing functionality (like the owners), but in addition it's missing some validation that this is a known package name...

src/discussions-trigger.ts
```diff
@@ -0,0 +1,175 @@
...
+    const discussionComments = await getCommentsForDiscussionNumber(discussion.number);
+    const previousComment = discussionComments.find(c => c.author.login === "typescript-bot");
+    if (previousComment) {
+        await client.mutate(createMutation<any>("updateDiscussionComment" as any, { body: message, commentId: previousComment.id }));
+    } else {
+        await client.mutate(createMutation<any>("addDiscussionComment" as any, { body: message, discussionId: discussion.node_id }));
+    }
+}
+
+async function addLabel(discussion: Discussion, labelName: string, description?: string) {
```
elibarzilay, line 111: Wouldn't it upset GH if we end up with up to thousands of labels?

But more importantly, it looks like they're project-global, so having these massive number of labels is going to make dealing with labels pretty difficult.  The least that this should do is use some prefix for these things, otherwise I can post a message with a `[Critical]` and confuse people and code...

src/discussions-trigger.ts
```diff
@@ -0,0 +1,175 @@
...
+
+    const existingLabel = await getLabelByName(labelName);
+    if (existingLabel && existingLabel.name === labelName) {
+        await client.mutate(createMutation<any>("addLabelsToLabelable" as any, { labelableId: discussion.node_id, labelIds: [existingLabel.id] }));
+    } else {
+        const dtRepoID = "MDEwOlJlcG9zaXRvcnk2MDkzMzE2";
+        // https://docs.github.com/en/graphql/reference/input-objects#createlabelinput
+        const color = crypto.randomBytes(Math.ceil(6/2)).toString("hex").slice(0,6);
```
elibarzilay, line 119: Won't it be easier to: `sha256(labelName).substring(0,6)`?

src/discussions-trigger.ts
```diff
@@ -0,0 +1,175 @@
...
+
+        const newLabel = await client.mutate(createMutation("createLabel" as any, { name: labelName, repositoryId: dtRepoID, color, description })) as any;
+        const newID = newLabel.data.label.id;
+        await client.mutate(createMutation<any>("addLabelsToLabelable" as any, { labelableId: discussion.node_id, labelIds: [newID] }));
```
elibarzilay, line 123: This line should be pulled out of the `if` since it's the same in both sides.

src/discussions-trigger.ts
```diff
@@ -0,0 +1,175 @@
...
+    }
+}
+
+async function addLabel(discussion: Discussion, labelName: string, description?: string) {
+
+    const existingLabel = await getLabelByName(labelName);
+    if (existingLabel && existingLabel.name === labelName) {
+        await client.mutate(createMutation<any>("addLabelsToLabelable" as any, { labelableId: discussion.node_id, labelIds: [existingLabel.id] }));
+    } else {
+        const dtRepoID = "MDEwOlJlcG9zaXRvcnk2MDkzMzE2";
```
elibarzilay, line 117: Better to add `id` to the `GetLabel` query and use that istead

src/discussions-trigger.ts
```diff
@@ -0,0 +1,175 @@
...
+}
+
+async function getCommentsForDiscussionNumber(number: number) {
+    const info = await client.query({
+        query: gql`
+query GetDiscussionComments($discussionNumber: Int!) {
+  repository(name: "DefinitelyTyped", owner: "DefinitelyTyped") {
+    name
+    discussion(number: $discussionNumber) {
+      comments(first: 100) {
```
elibarzilay, line 157: This should get them from the beginning of the thread, right?  So I think that it's better to add an `orderBy` (as in `all-open-prs-query.ts`).

elibarzilay (changes requested):
Lots of mostly little things...

src/discussions-trigger.ts
orta: Yeah, I've opted for "Pkg: " - I'd like something a bit further in the alphabet if you can thing of something better though? (emoji/non-ascii aren't counted in the sorting for labels)

src/graphql-client.ts
```diff
@@ -10,6 +10,7 @@ const apolloCfg = require("../apollo.config.js").client.service;
 const uri = apolloCfg.url;
 const headers = {
     ...apolloCfg.headers,
+    Accept: "application/vnd.github.bane-preview+json",
```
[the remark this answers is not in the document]
orta: Cool, agree, done!

package.json
```diff
@@ -12,9 +12,7 @@
   "bugs": {
     "url": "https://github.com/DefinitelyTyped/dt-mergebot/issues"
   },
-  "engines": {
-    "node": ">=14.16.1"
-  },
+
```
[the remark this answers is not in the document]
orta: Nope accident! Fixed

README.md
```diff
@@ -123,3 +123,16 @@ npm run update-all-fixtures
 
 Be careful with this, because PRs may now be in a different state e.g. it's now merged and it used to be a specific
 weird state.
+
+## Running with real webhooks
+
+You need a tool like [ngrok](https://ngrok.com) to expose a URL from the [webhooks section](https://github.com/DefinitelyTyped/DefinitelyTyped/settings/hooks/new) on DT.
```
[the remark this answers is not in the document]
orta: Agree, added to the getting started section above

Discussions-Trigger/sample.dat
```diff
@@ -0,0 +1,3 @@
+{
+    "name": "Azure"
+}
```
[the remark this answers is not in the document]
orta: dropped, aye

src/discussions-trigger.ts
```diff
@@ -0,0 +1,175 @@
...
+import { Discussion, DiscussionWebhook } from "./types/discussions";
+import { createMutation, client } from "./graphql-client";
+import fetch from "node-fetch";
+import { gql } from "@apollo/client/core";
+import * as crypto from "crypto";
+
+const reply = (context: Context, status: number, body: string) => {
+    context.res = { status, body };
+    context.log.info(`${body} [${status}]`);
+};
```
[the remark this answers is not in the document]
orta: Extracted 👍🏻  and done the same thing over on pr-trigger

src/discussions-trigger.ts
orta: It would! Doing so

src/discussions-trigger.ts
orta: Done

src/discussions-trigger.ts
orta: Not sure I get this one (unless you mean hardcode the number into the query?)

src/discussions-trigger.ts
orta: Thanks, good spot

src/discussions-trigger.ts
elibarzilay: I didn't see the `Pkg:` change, so leaving this open.  Also: I'm still worried about the first point: we have thousands of packages, and therefore we'd end up with thousands of labels, and I'm not sure how things would look like if/when we have that.

src/discussions-trigger.ts
elibarzilay: (didn't see the change, so leaving open.)

src/discussions-trigger.ts
elibarzilay: (didn't see the change, so leaving open.)

elibarzilay:
I did some tweaks (main one is negating the result of the verify function), and marked all of the resolved comments as such.

I could do some of these too, but you said that you did them so I'm guessing that maybe you forgot to push some more changes, and if I'll fix these I'll create a conflict nightmare for you -- so left them as-is.

src/discussions-trigger.ts
orta: In now

orta:
Alright, I'm gonna nurse this deploy

PR #423, opened 2021-09-06
Add a ping about 'hey you can do tests for this'

orta:
Adds a message hinting that you can skip waiting on a DT owner


src/comments.ts
```diff
@@ -120,6 +121,20 @@ export const WaitUntilMergeIsOK = (user: string, abbrOid: string, uri: string, m
...
+    tag: `wait-for-merge-offer-${abbrOid}`,
+    status: txt`
+        |:hourglass_flowing_sand: Hi @${user},
+        |
+        |It's been a few days since this PR was approved by ${approvalUsers.join(", ")}
+         and we're waiting on a DT maintainer to give a review.
+        |
+        |If you would like to short-circuit this wait, you can edit some of the [test files](${urls.testsTs})
+         in the package which verify how the \`.d.ts\` files work. This would allow the PR to be
+         merged by you or the DT module owners.`
```
elibarzilay, line 135: (I added a url constant and made the source more readable.)

elibarzilay (approved):
I like the idea, but a possible confusion is that people would expect to commit a test and be able to merge -- but in reality, the change would void the approvals so they'll need to wait for re-approvals.  Maybe mention this in the comment in some way to avoid the surprise (while not discouraging people from actually adding the tests)?

.gitignore
```diff
@@ -2,3 +2,4 @@
 /node_modules/
 /coverage/
 tsconfig.tsbuildinfo
+.DS_Store
```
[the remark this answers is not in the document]
orta: intended, this Mac is a fresh install

src/comments.ts
orta: Thanks - yeah, I'll add "after a re-review" and I think this is good then

orta:
Good from my side 👍🏻

src/comments.ts
elibarzilay: Yeah, that's a good addition

PR #425, opened 2021-09-09
Improve `RemindPeopleTheyCanUnblockPR`

elibarzilay:
When the CI is not passing, it seems weird to tell people that they can
short-circuit waiting for a maintainer.  But it'll be bad to drop the
comment altogether in this case: this will let people fix the CI and
only then bug them about adding a test, so they'll be much less
inclinded to do so.

Instead, have a different phrasing for the failing CI case.


elibarzilay:
Two comments:
* also changed "waiting on a maintainer" to "for a"
* not sure about "short-circuit the future maintainer approval"...

orta (approved):

PR #427, opened 2021-09-14
Improve `PingReviewersOther`

elibarzilay:
Don't say that "you're the only owner" if the author is not an existing
owner or a new one.


orta (approved):
Nice

PR #431, opened 2021-12-02
Update tslint "extends" requirement

sandersn:
formerly `dtslint/dt.json`, now `@definitelytyped/dtslint/dt.json`.

Also update a couple of baselines that were missed in earlier PRs.


src/_tests/fixtures/55210/mutations.json
```diff
@@ -7,5 +7,14 @@
...
       }
     }
+  },
```
sandersn, line 10: this and the change in result.json are the result of an earlier PR as far as I can tell.

src/_tests/fixtures/55741/derived.json
```diff
@@ -9,6 +9,7 @@
   "hasMergeConflict": false,
   "isFirstContribution": true,
   "tooManyFiles": false,
+  "hugeChange": false,
```
sandersn, line 12: same, from an earlier PR I think

orta (approved):
Looks good to me

elibarzilay:
(Bah.  I've had those changes in my directory, but I somehow missed adding them to that PR...)

PR #433, opened 2022-01-05
Mimic `git merge-base`

elibarzilay:
Github doesn't provide a merge-base thing, so previously the code used
"master" to compare against, which can easily fail.

Find a better commit to compare config contents against (config/owner
checks) using the list of commits in the PR.

Note that an alternative would be to use GH's `potentialMergeCommit`
(compare "master" as the old, and this as the new).  It takes a while
for it to be generated, but we already wait until `mergeable` is not
"UNKNOWN" which might mean that this is ready too.  But even with that,
it seems like a bad idea to rely on information from these temporary
merge branches.


orta (approved):
Looks good to me

PR #436, opened 2022-02-17
Put github-actions PRs into NMA

andrewbranch:


src/compute-pr-actions.ts
```diff
@@ -335,11 +335,15 @@ export function process(prInfo: BotResult,
...
-        && info.approverKind === "maintainer" && (info.staleness?.days ?? 0) > 3) {
+        && info.approverKind === "maintainer" && (info.staleness?.days ?? 0) > 3 && info.author !== "github-actions") {
         post(Comments.RemindPeopleTheyCanUnblockPR(info.author, info.approvedReviews.map(r => r.reviewer),
                                                    info.ciResult === "pass", headCommitAbbrOid));
     }
 
+    if (info.author === "github-actions") {
+        actions.projectColumn = "Needs Maintainer Action";
+    }
+
```
elibarzilay, line 346: It'll make it simpler & easier to follow if instead of doing this here it'll just test the author at the beginning of this section (before the `First-timers` comment), and it could also just `return actions` there which (a) makes the previous added test unnecessary, (b) also makes the staleness comment change not needed since it won't do the timeline actions anyway.  (And if there are more bots later then the same should apply to them too.)

And if these changes don't need the whole machinery of creating the welcome message, it can be done even before that which simplifies it further (since that would be the only change).

andrewbranch: I was concerned that if I returned too early, the logic to move cards into "done" or remove them from the board entirely wouldn’t fire. Did I misread that?

elibarzilay: Do you mean the `doneColumn` in the staleness thing?  -- Those are when the unmerged/abandoned/unreviewed timelines are done, and (I think that) you don't want to apply any of them to these PRs.

andrewbranch: Obviously the welcome comment is a little silly, but I do like the summary of changed packages, the debug info at the bottom, and owners getting pinged. Took the rest of your suggestions, thanks 👍

elibarzilay: Yeah, that makes sense (as well as keeping the labels too).  But one thing about the silliness, if it's kept, is that it will probably not be exercised since it's only for new contributors...?

andrewbranch: No, I inverted the `isFirstContribution` condition to make the nested ternary nicer

andrewbranch: The test reflects this since @github-actions is already a Contributor in the fixtures

PR #454, opened 2023-02-10
Add a proper handling of failures when executing the command

stof:
This ensures that a failure actually gets reported as a failure in the execution of the workflow on GitHub Actions


orta:
Thanks!

PR #456, opened 2023-02-13
Update github actions to their latest version

stof:
This avoids using the deprecated node 12.x runtime.


stof:
I'm not talking about the version of node installed with `setup-node`, but about the runtime used by the actions themselves: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

orta:
Great, makes sense to me - thanks

PR #461, opened 2023-04-12
Allow owners to unsubscribe from some PRs

### LINKED ISSUE #460: Stop pinging owners over and over for the same PR (allow unsubscribing)
Pings are necessary to make people aware of a new PR. If they don't take action, every following comment is already delivered to their notification **by default**. Continuous pinging doesn't do anything extra.

I want owners to have two options:

- do nothing (keep receiving updates)
- click unsubscribe (stop receiving updates)

The second option is not available because the bot keeps bugging everyone in nearly every comment.

fregante:
- Fixes #460 

I'm sure this is incomplete, I'm just showing the suggested path forward. If accepted I can complete it.

### DIFF
--- a/src/comments.ts
+++ b/src/comments.ts
@@ -60,7 +60,7 @@ export const SuggestTesting = deletedWhenNotPresent("suggest-testing", tag =>
 export const PingReviewers = (names: readonly string[], reviewLink: string) => ({
     tag: "pinging-reviewers",
     status: txt`
-        |🔔 ${names.map(n => `@${n}`).join(" ")} — please [review this PR](${reviewLink}) in the
+        |🔔 Pinging owners — please [review this PR](${reviewLink}) in the
          next few days. Be sure to explicitly select **\`Approve\`** or **\`Request Changes\`**
          in the GitHub UI so I know what's going on.`
 });
@@ -90,7 +90,7 @@ export const PingReviewersTooMany = (names: readonly string[]) => ({
 export const PingStaleReviewer = (reviewedAbbrOid: string, reviewers: string[]) => ({
     tag: `stale-ping-${sha256(reviewers.join("-")).substr(0, 6)}-${reviewedAbbrOid}`,
     status: txt`
-        |@${reviewers.join(", @")} Thank you for reviewing this PR! The author has pushed new
+        |**${reviewers.join(", ")}** Thank you for reviewing this PR! The author has pushed new
          commits since your last review. Could you take another look and submit a fresh review?`
 });
 
@@ -106,8 +106,7 @@ export const OfferSelfMerge = deletedWhenNotPresent("merge-offer", tag =>
             |> Ready to merge
             |
             |and I'll merge this PR almost instantly. Thanks for helping out! :heart:
-            |${otherOwners.length === 0 ? "" : `
-            |(${otherOwners.map(o => "@" + o).join(", ")}: you can do this too.)`}`}));
+            |${otherOwners.length === 0 ? "" : `(Any owner can do this too.)`}`}));
 
 export const WaitUntilMergeIsOK = (user: string, abbrOid: string, uri: string, mainCommentID: number | undefined) => ({
     // at most one reminder per update
@@ -128,7 +127,7 @@ export const RemindPeopleTheyCanUnblockPR = (user: string, approvalUsers: string
     status: txt`
         |:hourglass_flowing_sand: Hi @${user},
         |
-        |It's been a few days since this PR was approved by ${approvalUsers.join(", ")} and we're waiting for
+        |It's been a few days since this PR was approved and we're waiting for
          ${ciPassing ? `a DT maintainer to give a review`
                      : `you to fix the test failures and then for a maintainer approval`}.
         |


src/comments.ts
```diff
@@ -90,7 +90,7 @@ export const PingReviewersTooMany = (names: readonly string[]) => ({
 export const PingStaleReviewer = (reviewedAbbrOid: string, reviewers: string[]) => ({
     tag: `stale-ping-${sha256(reviewers.join("-")).substr(0, 6)}-${reviewedAbbrOid}`,
     status: txt`
-        |@${reviewers.join(", @")} Thank you for reviewing this PR! The author has pushed new
+        |**${reviewers.join(", ")}** Thank you for reviewing this PR! The author has pushed new
```
fregante, line 93: This is a way to mention people without actuating pinging them. It could be used anywhere.

src/comments.ts
```diff
@@ -128,7 +127,7 @@ export const RemindPeopleTheyCanUnblockPR = (user: string, approvalUsers: string
     status: txt`
         |:hourglass_flowing_sand: Hi @${user},
         |
-        |It's been a few days since this PR was approved by ${approvalUsers.join(", ")} and we're waiting for
+        |It's been a few days since this PR was approved and we're waiting for
```
fregante, line 130: Actually I think this was already non-pinging, so maybe this change isn't necessary

fregante (commented):
There are probably more messages where the ping should be dropped, like `StalenessComment`

src/comments.ts
```diff
@@ -60,7 +60,7 @@ export const SuggestTesting = deletedWhenNotPresent("suggest-testing", tag =>
 export const PingReviewers = (names: readonly string[], reviewLink: string) => ({
     tag: "pinging-reviewers",
     status: txt`
-        |🔔 ${names.map(n => `@${n}`).join(" ")} — please [review this PR](${reviewLink}) in the
+        |🔔 Pinging owners — please [review this PR](${reviewLink}) in the
```
jakebailey, line 63: I believe this is the initial comment on PRs, which should ping.

src/comments.ts
jakebailey: Correct, these are just usernames.

src/comments.ts
```diff
@@ -106,8 +106,7 @@ export const OfferSelfMerge = deletedWhenNotPresent("merge-offer", tag =>
             |> Ready to merge
             |
             |and I'll merge this PR almost instantly. Thanks for helping out! :heart:
-            |${otherOwners.length === 0 ? "" : `
-            |(${otherOwners.map(o => "@" + o).join(", ")}: you can do this too.)`}`}));
+            |${otherOwners.length === 0 ? "" : `(Any owner can do this too.)`}`}));
```
jakebailey, line 109: Instead of pinging nobody, this should probably ping those reviewers which are also owners.

src/comments.ts
fregante: Yes you're right, this was a mistake. 🙏

src/comments.ts
fregante: After the first mention in `PingReviewers`, every owner is already subscribed and will receive a notification for each commit and comment.

I think this is easier to reason about if you think `@user` as a `Subscribe @user to this PR` rather than `Ping @user`. Do you want to _subscribe_ them over and over?? That's the problematic part.

There should be one subscription, and then GitHub takes care of successive notifications already.

jakebailey: I disagree that it is "over and over" to remind everyone to actually merge the PR.

fregante: Mentions are not reminders, they're subscriptions. GitHub **already** reminds you after the subscription. A mention offers no improvement over _no mention_ if you're already subscribed.

jakebailey:
Most likely you're going to need to run something like this to fix the tests, which snapshot replies to PRs.

```console
$ AUTH_TOKEN=$(gh auth token) npm run update-all-fixtures
```

(But, after the comment threads are addressed, e.g. the "It's been a few days" does not to be changed)

src/comments.ts
jakebailey: I don't fully agree. Explicit mentions are delivered to a different email address than subscriptions, a fact that some people use to filter their inboxes. Not to mention the notification page, which splits them apart:

image

So, I think that there is still nuance here, moreso than the simple "is or isn't subscribed". Someone could reasonably use "mentioned" as being a higher priority than "subscribed".

But, I also don't have a strong opinion. But, I also have not seen anyone complain about this whole thing outside of your linked issue either.

fregante: Please believe when I say that **mentions are subscriptions**.

> Explicit mentions are delivered to a different email address than subscriptions

No.

Look at these two emails: one has a mention, the other is a followup notification. They have identical to/cc fields.

Further mentions or lack of mentions do no change that, unless it's a new "subscription" reason, like "assigned" or "review requested", in which case _that_ sticks around.

<table>
<tr>
 <td valign=top>
 <td valign=top> 
</table>


> Not to mention the notification page, which splits them apart

No. Same exact thing here. Once you've been mentioned, new notifications will continue to appear in the "Mentioned" filter.

… because, once again, GitHub thinks of "subscription reason", not "notification reason"



> But, I also don't have a strong opinion

💙 

> I also have not seen anyone complain

It looks me months to find this bot. Most people will just block the repo or just [asked to be removed](https://github.com/DefinitelyTyped/DefinitelyTyped/pulls?q=is%3Apr+is%3Aclosed+label%3AUnowned) altogether.

Heck I opened this issue in April 2023 and we're only discussing it now, which option is easier for people? 😃

fregante:
> the "It's been a few days" does not to be changed

No because that exclusively affects people who have unsubscribed, as shown in my other "review" comment.



> Most likely you're going to need to run something like this to fix the tests, which snapshot replies to PRs.

Snapshots updated 🫡 Tests pass locally 🟢 PR ready for merge 🚢

orta:
Just to generally chime in here, I'm not wild on these changes 

I think our biggest problem as DT maintainers is folks with real context not responding to these pings and removing the direct mentions in a bunch of places makes it more likely for them to be ignoring the API response. Like Jake, I have not heard people complain about this outside of the issue you reported and I'm not super convinced that we should be changing this.

fregante:
@orta please read the conversation we had so far. People will still be notified. This only affects people who have clicked Unsubscribe after the first notification.

src/comments.ts
jakebailey: > No. Same exact thing here. Once you've been mentioned, new notifications will continue to appear in the "Mentioned" filter.

From one of your pings:

Screenshot_20240127-045531

Then the email for the above message:

Screenshot_20240127-045708

Notice no "mentioned" the second time. It's possible that the merge event behaves inconsistently.

> Most people will just block the repo or just [asked to be removed](https://github.com/DefinitelyTyped/DefinitelyTyped/pulls?q=is%3Apr+is%3Aclosed+label%3AUnowned) altogether.

That's the wrong logical direction; maybe all annoyed people remove themselves, but not all people who remove themselves do so due to pings specifically.

fregante: @jakebailey did you unsubscribe from this thread between January 19 and January 26? I just posted screenshots that show the opposite behavior, so either you unsubscribed, or the behavior isn't really reliable.

Either way the question is simple, I won't continue to debate it:

- do you want to bother people who clicked Unsubscribe?

Because that's not great netiquette, also technically violating CAN-SPAM

jakebailey: > did you unsubscribe from this thread between January 19 and January 26?

No, I did not. Thinking about it, though

jakebailey: To be clear, I am not against the change, but just trying to explain that there is nuance in a workflow change like this.

I'm just not really understanding the antagonism here 😞

fregante:
It has been more than a year and this PR still has no reviews.

I'll bump it to the DT maintainer queue. Thank you for your patience, @jakebailey @sandersn @andrewbranch.

PR #468, opened 2023-05-18
Update for DT monorepo

sandersn:
The only code change is to switch to parsing package.json instead of a header from index.d.ts.
All the baselines have to change, though.


src/pr-info.ts
```diff
@@ -527,14 +519,18 @@ function downloadsToPopularityLevel(monthlyDownloads: number): PopularityLevel {
 }
 
 export async function getOwnersOfPackage(packageName: string, oid: string, fetchFile: typeof defaultFetchFile): Promise<string[] | null | Error> {
-    const indexDts = `${oid}:types/${packageName}/index.d.ts`;
-    const indexDtsContent = await fetchFile(indexDts, 10240); // grab at most 10k
-    if (indexDtsContent === undefined) return null;
-    let parsed: HeaderParser.Header;
+    const packageJson = `${oid}:types/${packageName}/package.json`;
+    const packageJsonContent = await fetchFile(packageJson, 10240); // grab at most 10k
```
jakebailey, line 523: I suspect we can drop this, now that it's not a huge declaration file that has to be parsed.

src/pr-info.ts
```diff
@@ -527,14 +519,18 @@ function downloadsToPopularityLevel(monthlyDownloads: number): PopularityLevel {
...
+    if (packageJsonContent === undefined) return null;
+    let packageJsonObj;
     try {
-        parsed = HeaderParser.parseHeaderOrFail(indexDtsContent);
+        packageJsonObj = JSON.parse(packageJsonContent);
     } catch (e) {
-        if (e instanceof Error) return new Error(`error parsing owners: ${e.message}`);
+        if (e instanceof Error) return new Error(`error parsing owners from package.json: ${e.message}`);
+    }
+    const parsed = HeaderParser.validatePackageJson(packageName, packageJsonObj, []);
```
jakebailey, line 531: I assume the error here is because this uses the new infra?

jakebailey: Which, metapoint, when does this repo get republished? Presumably we can change dt-tools and this doesn't get updated, which feels strange.

I sorta wonder how much of the mergebot could be moved to GHA...

PR #479, opened 2023-08-12
fix: enable cors for api routes

brc-dd:
dtmergebot's API used to have CORS enabled. v2 broke [orta/playground-dt-review](https://github.com/orta/playground-dt-review/pull/6)

Considering that the plugin is still shown in comments on the dt repo, it'd be better to enable CORS. Currently that "You can test the changes of this PR in the Playground." thing doesn't work.

If abuse is a concern, we can limit it to just `typescriptlang.org`

### DIFF
--- a/src/functions/api.ts
+++ b/src/functions/api.ts
@@ -3,7 +3,8 @@ import { getPRInfo } from "../queries/pr-query";
 const headers = {
     "Content-Type": "text/json",
     "Access-Control-Allow-Methods": "GET",
-    "Access-Control-Allow-Origin": "*",
+    "Access-Control-Allow-Origin": "https://www.typescriptlang.org",
+    "Vary": "Origin"
 };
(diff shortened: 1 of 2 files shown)


jakebailey:
@andrewbranch might know more here, though reading orta/playground-dt-review#6 I'm a little confused as to if this public API is actually meant to be consumed given we seemingly changed the URL during a re-deployment a while ago (to fix something) and nobody noticed?

andrewbranch:
I wasn’t aware of this trigger, but I assume Orta built it for this purpose.

> If abuse is a concern, we can limit it to just typescriptlang.org

Let’s go ahead and do this—more that I don’t want to imply ongoing support for this endpoint as public API, though I’m fine with it being used while it’s not a maintenance burden.

Thanks!

andrewbranch (approved):

PR #482, opened 2023-10-18
Use old header parser when package.json doesn't look new

jakebailey:
https://github.com/DefinitelyTyped/dt-mergebot/commit/6b659eb7177af948e7e649bb65266bd16d7ef7ea#diff-2b507820ea5da67eac1be3f3c554e982520499201e3b667d69cea285a5ed59d0 and the recent changes are making the bot unhappy when detecting owner changes.

Instead of trying to figure out how to find the real merge commit and to the comparison, we can instead just pull the header code back in and use it conditionally. This will work better for older PRs which still contain `index.d.ts` and so won't misparse as empty owners.

~Rerunning tests doesn't seem to change anything. I had to recreate fixtures to actually get something to show my fix so I'm not sure what's going on there. I guess I need `npm run update-all-fixtures`?~

I reverted the tests to before the DT change and then reran tests here.

This diff is whack; https://github.com/DefinitelyTyped/dt-mergebot/compare/e49a4e17d1c5440f7e5fc37a0365ca5a305faa33...jakebailey:dt-mergebot:fix-diffing shows the difference between this PR and the change before the DT change, which should be much more reasonable.


package.json
```diff
@@ -15,6 +15,7 @@
   "license": "MIT",
   "dependencies": {
     "@apollo/client": "latest",
+    "@definitelytyped/old-header-parser": "npm:@definitelytyped/header-parser@0.0.178",
```
jakebailey, line 18: I named it something different to force someone to think twice about using it. But I can just use a regular dep.

src/pr-info.ts
```diff
@@ -520,12 +521,28 @@ function downloadsToPopularityLevel(monthlyDownloads: number): PopularityLevel {
 export async function getOwnersOfPackage(packageName: string, oid: string, fetchFile: typeof defaultFetchFile): Promise<string[] | null | Error> {
     const packageJson = `${oid}:types/${packageName}/package.json`;
     const packageJsonContent = await fetchFile(packageJson, 10240); // grab at most 10k
-    if (packageJsonContent === undefined) return null;
     let packageJsonObj;
-    try {
-        packageJsonObj = JSON.parse(packageJsonContent);
-    } catch (e) {
-        if (e instanceof Error) return new Error(`error parsing owners from package.json: ${e.message}`);
+    if (packageJsonContent !== undefined) {
```
jakebailey, line 525: Alternatively, I can try the header _first_ and only break out if it fails to parse.

sandersn (approved):
I'll add a future TODO to remove this in a week or so.

jakebailey:
I don't think we should remove this for a long time, people may undraft old PRs and we shouldn't choke on them.

PR #484, opened 2023-12-13
Remove tslint

sandersn:
And do not add checking for eslint. Not sure that's the right choice.

OTHER_FILES.txt also isn't checked anymore. Also there are some unused link constants.


sandersn:
@JoshuaKGoldberg you are probably interested in this too.

JoshuaKGoldberg:
👍 this is of interest to me, thanks for the ping! I don't have anything to add but I am very happy about it. 🙂

jakebailey (approved):
LGTM but I think we'll have to remove tslint.json from the repo first as we'll now treat them as infrastructure and nobody's PRs will be mergable.

jakebailey:
Or maybe we are forced to leave it, just return undefined like the other files?

sandersn:
I'd rather have a PR ready to remove them from DT first, then merge both the dt-tools PR and this PR. At least, I don't think there's any point in allowing-but-not-checking them like .npmignore et al.

PR #485, opened 2024-01-31
Handle workflow reruns for detecting CI failures

jakebailey:
Before this PR, we would just take the first check per commit and treat that as CI. But, when you close/reopen a PR or rerun a workflow, that new check is _appended_ to the list of checks.

Rather than always taking the first check, take the latest check per workflow, then check _all_ checks for success.

See https://github.com/DefinitelyTyped/dt-mergebot/commit/2892c8fdd1b8de3ce518660d7f54caac88c40690 for the actual diff without the noise.

To do this, I had to update the graphql schema, which was super annoying given our apollo dep is very out of date.


andrewbranch:
I think this looks good, but there are some snapshot failures.

jakebailey:
Yep, just fixed those; two old PRs that actually had failing checks when merged!

andrewbranch (approved):

PR #493, opened 2024-04-11
chore: Replace most usages of "which" with "that"

stephenwade:
I noticed some bad grammar in the DefinitelyTyped repo comments, so I've created this PR to fix it.

Most places in this code that use "which" should use "that". "that" is used in restrictive clauses, and "which" is used in nonrestrictive clauses.

More info about the grammar: https://www.dictionary.com/e/that-vs-which/


jakebailey:
Not sure why this didn't fail before, but all of the baselines that include this text have to be updated to be passing.

stephenwade:
@jakebailey I updated the baselines with `npm test -- -u`.

jakebailey:
I think you missed merging from master before that, sorry!

stephenwade:
@jakebailey How's this?

jakebailey (approved):

PR #508, opened 2024-06-26
Switch to pnpm

jakebailey:
In prep for moving to dt-tools; this should now be safe now that we bundle before uploading to Azure Functions.


.github/workflows/deploy.yml
```diff
@@ -28,14 +28,19 @@ jobs:
...
+          pnpm install
+          pnpm run build --if-present
+          pnpm run test --if-present
+
+      - name: Bundle
+        run: pnpm run bundle
 
-      - name: Build zip
-        run: npm run bundle
+      - name: Check bundle
```
sandersn, line 42: I'm guessing this isn't related to the pnpm check? Seems like we just weren't doing it on deploy before?

.github/workflows/test.yml
```diff
@@ -29,12 +30,9 @@ jobs:
...
-    - run: |
-        export BOT_AUTH_TOKEN=secret
-        for i in dist/functions/*.js; do 
-          node $i;
-        done
+    - uses: pnpm/action-setup@fe02b34f77f8bc703788d5817da081398fad5dd2 # v4.0.0
+    - run: pnpm install
+    - run: pnpm run build
+    - run: pnpm test
+    - run: pnpm run bundle
```
sandersn, line 37: Why do we have to bundle now when we didn't before?

sandersn (approved):
I think it's ready to go but I have a couple of questions.

.github/workflows/deploy.yml
jakebailey: Yes, correct; an oversight in my last PR to bundle instead of uploading `node_modules`.

.github/workflows/test.yml
jakebailey: This was also an oversight in my last PR; we used to in this check test the compiled code, but that doesn't really matter, we want to verify that the bundle we're going to send to Azure even runs, just as an extra safety check.

